PR #75, opened 2024-06-22
feat(ingestion_pipeline): optional error filtering and logging

### LINKED ISSUE #64: Improved error handling
Currently, depending on the transformer used, if any node has an error, the pipeline will fail early. A benefit is that errors are always bubbled up to the end of the pipeline. To make it configurable and modular, we could add a step that enables for instance to filter out/log errors instead.

For instance:
```rust
IngestionPipeline::from_loader(FileLoader::new(".").with_extensions(&["rs"]))
        .then(MetadataQACode::new(openai_client.clone()))
        .handle_errors(ErrorHandling::FilterAndLog)
        .run().await?
```

This allows fine grained control per step on errors without breaking the readability of the chain.

### LINKED ISSUE #73: Stream inspection
When building you might want to inspect or log nodes that come through. This is technically possible with a transformer. But perhaps we can make the api a bit nicer, like:

```rust
IngestionPipeline::from_loader(FileLoader::new(".").with_extensions(&["rs"]))
        .then(MetadataQACode::new(openai_client.clone()))
        .inspect_err()
        .inspect_ok()
        .inspect()
        .run().await?
```

This is similar to #64

timonv:
Closes #73 and closes #64

Introduces various methods for stream inspection and skipping errors at a desired stage in the stream.

### DIFF
--- a/swiftide/src/ingestion/ingestion_pipeline.rs
+++ b/swiftide/src/ingestion/ingestion_pipeline.rs
@@ -238,6 +238,56 @@ impl IngestionPipeline {
         self
     }
 
+    // Silently filters out errors encountered by the pipeline.
+    //
+    // This method filters out errors encountered by the pipeline, preventing them from bubbling up and terminating the stream.
+    // Note that errors are not logged.
+    pub fn filter_errors(mut self) -> Self {
+        self.stream = self
+            .stream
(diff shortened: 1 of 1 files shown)


tinco:
Nice I can get out of the feet with this.

PR #121, opened 2024-07-02
chore: update CHANGELOG.md

eltociear:
continous -> continuous


timonv:
Thank you for contributing! Unfortunately the changelog is auto generated and can't be fixed without rewriting history.

PR #122, opened 2024-07-02
fix(chunkcode): use correct chunksizes

timonv:


swiftide/src/integrations/treesitter/splitter.rs
```diff
@@ -109,26 +109,50 @@ impl CodeSplitter {
...
+
+                if next_child_size > self.max_bytes() {
+                    let mut sub_chunks =
+                        self.chunk_node(child, source, last_end, Some(current_chunk));
+                    current_chunk = sub_chunks.pop().unwrap_or(String::new());
+                    new_chunks.extend(sub_chunks);
+                } else {
+                    // NOTE: if the current chunk was smaller than then the min_bytes, then it is discarded here
+                    if !current_chunk.is_empty() && current_chunk.len() > self.min_bytes() {
+                        new_chunks.push(current_chunk);
```
tinco, line 148: This is now the only remaining location where an undersized `current_chunk` might be pushed onto `new_chunks`. If we want to eliminate this case as well I think we need to give up on this tree traversal thing and just flatten the whole thing only walking the leave nodes.

timonv: Treesitter has bidirectional walking. text_splitter uses this in an iterator to fix it instead. It's a very rusty implementation.

swiftide/src/integrations/treesitter/splitter.rs
```diff
@@ -109,26 +109,46 @@ impl CodeSplitter {
...
-            if child.end_byte() - child.start_byte() > self.max_bytes() {
-                // Child is too big, recursively chunk the child
-                if !current_chunk.is_empty() && current_chunk.len() > self.min_bytes() {
-                    new_chunks.push(current_chunk);
+            assert!(
+                current_chunk.len() <= self.max_bytes(),
+                "Chunk too big: {} > {}",
+                current_chunk.len(),
+                self.max_bytes()
+            );
```
timonv, line 128: How do assertions work outside of tests? Do they log if the condition passes when debug assertions are enabled?

tinco: No that's `debug_assert`, regular `assert` always cause a panic, both in debug and release builds. It's a technique used in high reliability projects like databases and operating systems to ensure invariants hold. Of course in Rust there's no risk of memory corruption so `assert` might be less useful outside of `unsafe` blocks. I'm fine with removing them if you think it's unnecessary.

swiftide/src/integrations/treesitter/splitter.rs
tinco: I'm not sure what you mean by that, do you mean you've got a solution in mind?

swiftide/src/integrations/treesitter/splitter.rs
timonv: I think you have a point that it's good to have a double check, it really shouldn't ever happen. Are you sure you want to panic? It would break any running pipeline. You could also debug assert or log an error, or both.

swiftide/src/integrations/treesitter/splitter.rs
timonv: I think it's a solvable problem by using the walker instead of a recursive for loop. But I don't think it's really that much of an easy to reimplement the whole thing right now. https://github.com/benbrandt/text-splitter/blob/main/src/splitter/code.rs#L194 is a good example. Personally I'd just let it be and get this merged 👯

tinco (approved):

Issue #62, opened 2024-06-21
Named / multiple vector support for Qdrant


timonv:
The current Qdrant implementation by default only supports a single vector. By supporting named vectors we enable two highly desired features:

* **Different metadata embedding modes** Currently we add the metadata to the chunk and then embed. Ideally we can also choose not to, or do both, and create and add vectors for each field. This might be a requirement for this issue.
* **Sparse vector support** Sparse vectors are great for lightweight hybrid search looking for more exact matches. 

See https://qdrant.tech/documentation/concepts/collections/#collection-with-multiple-vectors.

pwalski:
Would such implementation be ok:

1. Introduce public `transformers::embed::EmbedMode { WithMetadata, WithoutMetadata, Both }` type.
2. Update `transformers::embed::Embed` by creating a Builder (optionally replacing `with_concurrency` with `concurrency` derive_builder generated setter), and adding `embed_mode` field/builder setter with a default value `WithMetadata`.
3. Update `ingestion::IngestionNode` to have `vectors: Option<HashMap<String,<Vec<f32>>>>` instead of a `vector` field (it will do for now, but might require some `VectorType` to support Sparse vectors)
4. Make `impl BatchableTransformer for Embed::batch_transform` to create embeddable chunks depending on selected `EmbedMode` and add embeddings vectors to the map.
5. Make `impl Qdrant::create_index_if_not_exists` to create `qdrant_client::qdrant::CreateCollection` struct with `vectors_config` field of `Config::ParamsMap` value containing named vectors. Its value would depend on selected `EmbedMode`. There could be a default vec named _"WithMetadata"_, or _"WithoutMetadata"_, or both vectors. Their length would come from `Qdrant::vector_size` (but it could as well come from `impl EmbeddingModel` - if I understand correctly the length of dense vecs depends on selected embed model).
6. Tests/examples/etc.

?

Will it be enough to configure named vectors using only an `EmbedMode` field in `Embed` struct, or will users need to configure e.g. a different `Distance` for each vector, so introduction of some `vector_config` field in `Qdrant` builder will be necessary?
It could not even give user control over vector names, but instead it could have some dedicated config types for each `EmbedMode`.


Disclaimer: after a short break from Rust I need to do some tasks as an reminder. I am unfamiliar with RAG pipelines, so I do not know e.g. how much control over vectors users may need.

timonv:
Hi Pwalski, thanks for being interested and offering your help! Sounds really good. 

[...]

Issue #63, opened 2024-06-21
Sparse vector support


timonv:
Depends on #62

Sparse vectors are a great solution to do hybrid search on more exact matches without having multiple storage backends. 

There's two major sides to this feature:
* Converting dense vectors to sparse vectors (i.e. with SPLADE)
* Add support to the qdrant integration

Conceptually (if possible), it would be great if this feature, for the user, is a matter of configuration and doesn't require changes to the actual pipeline.

See also https://weaviate.io/blog/hybrid-search-explained

PR #123, opened 2024-07-03
Named vectors support

pwalski:
Resolves: #62 

I might backpedal from `EmbeddableType` idea - I used it to:
- to detect in Qdrant impl if arriving single vector is a default vector (for `EmbedMode::SingleWithMetadata`) - finally I opted to just store any single vector as a default vector
- to get a name for vector for chunk with metadata  and chunk only without necessity for user to provider this name

It could be replaced with some `const` names, like the ones I added for `Metadata` implementations.

Remaining ToDo:
- fix `EmbedMode::Both`
- tests
- docs


swiftide/src/transformers/metadata_qa_code.rs
```diff
@@ -7,6 +7,8 @@ use anyhow::Result;
 use async_trait::async_trait;
 use indoc::indoc;
 
+pub const NAME: &str = "Questions and Answers (code)";
```
pwalski, line 10: 1. I wanted to store somewhere vector/metadata names
2. QA Metadata Code and Text had duplicated names. I changed them to avoid problems downstream.

timonv:
Look good already, nice! Maybe there's a way to clean up the qdrant configuration. If you always need an EmbeddableType and sometimes a Vector config, you could also have the function signature like:

```rust
    pub fn with_vector(
        mut self,
        vector: impl Into<VectorConfig>,
    ) 
  ```
  
  And then implent `Into<VectorConfig>` for embedabletype. One requirement I do have is that if needed, qdrant can be utilized fully. What happens if a vector config is provided, but a custom client is build? 

  What I really like about your implementation is that it also sets swiftide up to infer the schema from the pipeline at a later stage.

**Comment by pwalski** [comment]
> One requirement I do have is that if needed, qdrant can be utilized fully. What happens if a vector config is provided, but a custom client is build?

I do not understand. 
`Qdrant` client implementation from this library limits user's ability to configure _Vectors_ by providing only few config options in `VectorConfigBuilder` (name, size, distance). New `Qdrant` client implementation would need to come with new implementation of configuring Vectors, so it does not matter if current `VectorConfigBuilder` supports only few config options.
The same goes with other `Qdrant` config options. `QdrantBuilder` provides only a subset of  `qdrant_client::Qdrant` config options.

**Comment by pwalski** [comment]
> One requirement I do have is that if needed, qdrant can be utilized fully. What happens if a vector config is provided, but a custom client is build?

I do not understand. `Qdrant` client implementation from this library limits user's ability to configure _Vectors_ by providing only few config options in `VectorConfigBuilder` (name, size, distance). New `Qdrant` client implementation would need to come with new implementation of configuring Vectors.
The same goes with other `Qdrant` config options. `QdrantBuilder` provides only a subset of  `qdrant_client::Qdrant` config options.

**Comment by timonv** [comment]
> > One requirement I do have is that if needed, qdrant can be utilized fully. What happens if a vector config is provided, but a custom client is build?
> 
> I do not understand. `Qdrant` client implementation from this library limits user's ability to configure _Vectors_ by providing only few config options in `VectorConfigBuilder` (name, size, distance). New `Qdrant` client implementation would need to come with new implementation of configuring Vectors. The same goes with other `Qdrant` config options. `QdrantBuilder` provides only a subset of `qdrant_client::Qdrant` config options.

Exactly, I don't see the need to replace / reinvent full apis from other libraries that already work. Most (if not all) integrations support setting whatever it wraps manually, allowing for more fine grained configuration. So we provide some basic stuff for sensible defaults, otherwise it's RTM on the integrations' documentation.

I thought the current setup might go wrong when overriding the client and accidentally also providing vector configuration.

File: examples/store_multiple_vectors.rs [hunk 2]
```diff
@@ -0,0 +1,72 @@
...
+                .batch_size(50)
+                .vector_size(1536)
+                .collection_name("swiftide-multi-vectors")
+                .with_vector(EmbeddableType::Chunk)
+                .with_vector(EmbeddableType::Metadata(metadata_qa_text::NAME.into()))
+                .with_vector(EmbeddableType::Metadata(metadata_summary::NAME.into()))
+                .with_vector(
+                    VectorConfig::builder()
+                        .embeddable_type(EmbeddableType::Metadata(metadata_title::NAME.into()))
+                        .distance(Distance::Manhattan)
```
pwalski, line 63: I left it like this to serve as an example of vector distance config.

swiftide/src/indexing/node.rs
```diff
@@ -39,10 +40,12 @@ pub struct Node {
     pub path: PathBuf,
     /// Data chunk contained in the node.
     pub chunk: String,
-    /// Optional vector representation of the data chunk.
-    pub vector: Option<Vec<f32>>,
+    /// Optional vector representation of embedded data.
+    pub vectors: Option<HashMap<EmbeddableType, Vec<f32>>>,
     /// Metadata associated with the node.
-    pub metadata: HashMap<String, String>,
+    pub metadata: BTreeMap<String, String>,
```
pwalski, line 46: I change it to `BTreeMap` to avoid random order in embeddables combining metadata with chunk, and to make testing a bit easier.

swiftide/src/indexing/node.rs
```diff
@@ -57,9 +60,16 @@ impl Debug for Node {
...
             .field(
-                "vector",
-                &self.vector.as_ref().map(|v| format!("[{}]", v.len())),
+                "vectors",
+                &self
+                    .vectors
+                    .iter()
+                    .map(HashMap::iter)
+                    .flatten()
+                    .map(|(embed_type, vec)| format!("'{embed_type}': {}", vec.len()))
```
pwalski, line 69: Logs will be a bit more verbose - instead of vector length it will be _metadata1: len1, metadata2: len2_

swiftide/src/indexing/node.rs
```diff
@@ -118,3 +150,31 @@ impl Hash for Node {
...
+#[derive(Clone, Default, Serialize, Deserialize, PartialEq, Eq, Hash, strum_macros::Display)]
+pub enum EmbeddableType {
+    #[default]
+    /// Embeddable created from Chunk of data combined with Metadata.
+    Combined,
+    /// Embeddable created from Chunk of data only.
+    Chunk,
+    /// Embeddable created from Metadata.
+    /// String stores Metadata name.
+    #[strum(to_string = "Metadata: {0}")]
```
pwalski, line 178: Vector created for metadata embeddable will have name created from `EmbeddableType::to_string`.

swiftide/src/integrations/qdrant/mod.rs
```diff
@@ -116,6 +153,26 @@ impl QdrantBuilder {
...
+    /// When not configured Pipeline by default configures vector only for [EmbeddableType::Combined]
+    /// Default config is enough when [crate::indexing::Pipeline::with_embed_mode] is not set
+    /// or when the value is set to [crate::indexing::EmbedMode::SingleWithMetadata].
+    pub fn with_vector(mut self, vector: impl Into<VectorConfig>) -> QdrantBuilder {
+        if self.vectors.is_none() {
+            self = self.vectors(Default::default());
+        }
+        let vector = vector.into();
+        if let Some(vectors) = self.vectors.as_mut() {
+            vectors.insert(vector.embeddable_type.clone(), vector);
```
pwalski, line 168: Silently overrides vector.
The library has no logger, but it would deserve at least log msg.

examples/store_multiple_vectors.rs
pwalski: A disclaimer comment.
Resolving it to calm down PR merging rules.

swiftide/src/indexing/node.rs
pwalski: A disclaimer comment.
Resolving it to calm down PR merging rules.

swiftide/src/indexing/node.rs
pwalski: A disclaimer comment.
Resolving it to calm down PR merging rules.

swiftide/src/integrations/qdrant/mod.rs
pwalski: A disclaimer comment.
Resolving it to calm down PR merging rules.

swiftide/src/indexing/node.rs
pwalski: A disclaimer comment.
Resolving it to calm down PR merging rules.

swiftide/src/transformers/metadata_qa_code.rs
pwalski: A disclaimer comment.
Resolving it to calm down PR merging rules.

swiftide/src/transformers/embed.rs
```diff
@@ -64,28 +67,221 @@ impl BatchableTransformer for Embed {
...
+            Err(err) => return IndexingStream::iter(Err(err)),
+        };
+
+        IndexingStream::iter(nodes.into_iter().map(move |mut node| {
+            let Some(embedding_keys) = embeddings_keys_groups.pop_front() else {
+                bail!("Missing embedding data");
+            };
+            let remaining_embeddings = embeddings.split_off(embedding_keys.len());
+            let embedding_values = embeddings.clone();
+            embeddings = remaining_embeddings;
```
timonv, line 101: You're cloning (large) vectors and then reassigning the previous value to what was cloned. Maybe there is an elegant way of solving this with drain. At the very least the clone should be avoidable.

timonv (changes requested):
Solid work! I have a couple of comments and some suggestions.

I think EmbeddedType could be more apptly named as `EmbeddedField` or `EmbedField`, as that seems to be more what it is.

Maybe there's also a way to clean up the api for `.with_vector`, `.with_vector(EmbeddableType::Metadata(metadata_summary::NAME.into()))`, do you have any ideas?

On a more practical note, this looks _very_ close to merging and I would love to get it in for 0.6! Do you have any social handles I can mention on release? 🎉

swiftide/src/integrations/qdrant/indexing_node.rs
```diff
@@ -49,11 +49,117 @@ impl TryInto<qdrant::PointStruct> for Node {
...
-        ))
+        Ok(qdrant::PointStruct::new(id, vectors, payload))
+    }
+}
+
+fn try_create_vectors(vectors: HashMap<EmbeddableType, Vec<f32>>) -> Result<qdrant::Vectors> {
+    if vectors.is_empty() {
+        bail!("Node with empty vectors")
+    } else if vectors.len() == 1 {
+        let vector = vectors.into_values().next().expect("Node has vector entry");
```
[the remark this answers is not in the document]
pwalski: I did just panic on `next()` because it should never happen thanks to `vectors.len() == 1` condition. I will change it to `bail!` because it looks wrong.

swiftide/src/transformers/embed.rs
```diff
@@ -64,28 +67,221 @@ impl BatchableTransformer for Embed {
...
+        let mut embeddings = match self.embed_model.embed(embeddables_data).await {
+            Ok(embeddngs) => VecDeque::from(embeddngs),
+            Err(err) => return IndexingStream::iter(Err(err)),
+        };
+
+        IndexingStream::iter(nodes.into_iter().map(move |mut node| {
+            let Some(embedding_keys) = embeddings_keys_groups.pop_front() else {
+                bail!("Missing embedding data");
+            };
+            let remaining_embeddings = embeddings.split_off(embedding_keys.len());
```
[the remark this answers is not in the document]
pwalski: `split_off(n)` removes elements starting from `n` index, so `embeddings` and `embedding_keys` will have a same length afterwards.

swiftide/src/transformers/embed.rs
pwalski: Fail. `embeddings` is a deque, so I changed it to just `pop_front()` while iterating over embedding keys.

pwalski:
>  Do you have any social handles I can mention on release?

I do not have any actively updated profiles.

> Maybe there's also a way to clean up the api for `.with_vector`, `.with_vector(EmbeddableType::Metadata(metadata_summary::NAME.into()))`, do you have any ideas?

These `NAME` fields got created because I had no idea how to configure vectors and share info about metadata/embedding/vector name.
The fact user needs to specify same Metadata once as a `Transformer` and once as a `VectorConfig` name is already a bit awkward.

The quoted line could be cleaned up a bit by providing following impl for `EmbeddedField`:
```rs

impl From<&str> for EmbeddedField {
    fn from(value: &str) -> Self {
        Self::Metadata(value.into())
    }
}
```

The problem is I am not sure how long such `From` impl would hold. It could become a problem even during implementation of Sparse vectors https://github.com/bosun-ai/swiftide/issues/63.
That task will introduce another usecase for named vectors so it could be a good opportunity to make some updates to vectors config.

swiftide/src/integrations/qdrant/mod.rs
pwalski: Done.

timonv:
> > Do you have any social handles I can mention on release?
> 
> I do not have any actively updated profiles.
> 
> > Maybe there's also a way to clean up the api for `.with_vector`, `.with_vector(EmbeddableType::Metadata(metadata_summary::NAME.into()))`, do you have any ideas?
> 
> These `NAME` fields got created because I had no idea how to configure Metadata and share info about metadata/embedding name with vector config. The fact user needs to specify same Metadata once as a `Transformer` and once as a `VectorConfig` name is already a bit awkward.
> 
> The quoted line could be cleaned up a bit by providing following impl for `EmbeddedField`:
> 
> ```rust
> impl From<&str> for EmbeddedField {
>     fn from(value: &str) -> Self {
>         Self::Metadata(value.into())
>     }
> }
> ```
> 
> The problem is I am not sure how long such `From` impl would hold. It could become a problem even during implementation of Sparse vectors #63. That task will introduce another usecase for named vectors so it could be a good opportunity to make some updates to vectors config.
> 
> Making `EmbeddedField` generic over some `T: Into<String> + Display` would cause propagation of generic type across codebase. Making `EmbeddedField::Metadata` accept `&str` would require usage of a lifetime, which creates same problem as generic.

Fully agree, I don't see a direct option here either. I have hunch that when we get further down the line, a nicer model will pop up. I.e. a pipeline could infer a schema.

Feel it's ready to merge? 0.6 will probably be released this week. Need to update and rewrite a bunch of documentation on swiftide.rs for it, it was setup a bit in a hurry.

pwalski:
> Feel it's ready to merge?

Yes.
There are issues like the fact `with_embed_mode()` needs to be placed before `then(impl Transformer)`, and multiple `then_store_with()` do not work, but it is not related directly to this PR.

timonv:
Multiple then store with don’t work? That’s a bug

timonv (approved):

Issue #138, opened 2024-07-10
Inability to store some of `EmbeddedField`s


pwalski:
**Describe the bug**
Named vectors support was implemented with an assumption that every `EmbeddedField` should be stored as an _Qdrant_ named vector.

**To Reproduce**
Removing any of `with_vector` calls from `store_multiple_vectors.rs` example will result with an empty Qdrant collection.

**Expected behavior**
Removing vector from `Qdrant` storage config of above example should result with skipping suitable `EmbeddedField` vector.

PR #139, opened 2024-07-10
fix: Inability to store only some of `EmbeddedField`s

pwalski:
Fixes: #138


timonv (dismissed):
Thanks @pwalski, looks great!

PR #145, opened 2024-07-11
feat(prompts): add prompts as first class citizens

timonv:
Adds Prompts as first class citizens. This is a breaking change as SimplePrompt with just a a `&str` is no longer allowed.

This introduces `Prompt` and `PromptTemplate`. A template uses jinja style templating build on tera. Templates can be converted into prompts, and have context added. A prompt is then send to something that prompts, i.e. openai or bedrock.

Additional prompts can be added either compiled or as one-offs. Additionally, it's perfectly fine to prompt with just a string as well, just provide an `.into()`.

For future development, some LLMs really benefit from system prompts, which this would enable. For the query pipeline we can also take a much more structured approach with composed templates and conditionals.

### DIFF
--- a/swiftide/src/integrations/aws_bedrock/simple_prompt.rs
+++ b/swiftide/src/integrations/aws_bedrock/simple_prompt.rs
@@ -1,4 +1,4 @@
-use crate::SimplePrompt;
+use crate::{prompt::Prompt, SimplePrompt};
 use anyhow::Result;
 use async_trait::async_trait;
 use aws_sdk_bedrockruntime::primitives::Blob;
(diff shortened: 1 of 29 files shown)


swiftide/src/integrations/aws_bedrock/simple_prompt.rs
```diff
@@ -53,7 +53,7 @@ mod test {
             .build()
             .unwrap();
 
-        let response = bedrock.prompt("Hello").await.unwrap();
+        let response = bedrock.prompt(&"Hello".into()).await.unwrap();
```
timonv, line 56: There has to be a nice way to not have this breaking change and just allow &'static str. `impl` does not work with automock and trait bounds in traits are meh.

PR #271, opened 2024-09-04
feat(query): Add custom search strategies for qdrant and lancedb

### LINKED ISSUE #259: Custom queries in query pipeline
Currently queries in the query pipeline can only be customized by providing a closure for the default strategy. Both lancedb and qdrant have involved boilerplate for extracting the document, and both provide a relatively simple builder for queries.

As a solution with a nicer api, we could provide a `Custom` search strategy, such that only the query has to be constructed manually and all the rest is done as is.

i.e.

```rust
struct CustomSearchStrategy<T> {
  query: T
}

impl<T> CustomSearchStrategy<T> {
  pub fn custom_query(&self) -> &T {
    &self.query
  }

  pub fn from_query(query: T> -> Self<T> {
    Self { query }
  }
}
    
impl SearchStrategy for CustomSearchStrategy {}


impl Retrieve<CustomSearchStrategy<QueryPointsBuilder>> for Qdrant {
  async fn retrieve(strategy: CustomSearchStrategy, query: Query) {
    let query_result = strategy.custom_query();
    # ... transform into documents
  }
}
```

timonv:
Tests & docs pending.

Closes #259

### DIFF
--- a/swiftide-core/src/query_traits.rs
+++ b/swiftide-core/src/query_traits.rs
@@ -32,7 +32,7 @@ where
 }
 
 /// A search strategy for the query pipeline
-pub trait SearchStrategy: Clone + Send + Sync + Default {}
+pub trait SearchStrategy: Clone + Send + Sync {}
 
 /// Can retrieve documents given a SearchStrategy
 #[async_trait]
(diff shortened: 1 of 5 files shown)


timonv:
Closing for now. The problem is we need to make a lot of assumptions on the underlying schema, which feels like it defeats the purpose of a custom strategy.

Perhaps in the future we have a better idea on what makes sense. In the mean time, the default strategy supports closures and that works fine.

PR #273, opened 2024-09-05
docs(README): Fix date

dzvon:
I suppose this should be 09-02.


timonv:
Hahaha, sharply spotted! Thank you. <3

PR #304, opened 2024-09-17
docs: fix broken documentation links and other cargo doc warnings

tinco:
Running `cargo doc --all-features` resulted in a lot of warnings. We should probably add this to the CI/CD pipeline to prevent it from degrading again.

### DIFF
--- a/swiftide-core/src/node.rs
+++ b/swiftide-core/src/node.rs
@@ -224,8 +224,6 @@ impl<T: Into<String>> From<T> for Node {
 }
 
 /// Embed mode of the pipeline.
-///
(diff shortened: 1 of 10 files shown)


timonv:
Nice! Do the links work when opening cargo doc? Afaik there's an 'issue' with not being able to link to other crates in the workspace.

timonv (approved):

tinco:
> Nice! Do the links work when opening cargo doc? Afaik there's an 'issue' with not being able to link to other crates in the workspace.

Yeah, it's not really a bug but more of a missing feature. There's some solutions out there but they're basically hacks. The best solution is just to either not link, or to do a full external link manually.

PR #309, opened 2024-09-21
feat(tree-sitter): Add support for Java 22

devsprint:


tinco:
Very cool! I suppose we don't have any tests for the refs/defs module.

devsprint:
I have added a basic test in `outliner.rs`.  The linter reported issue I guess that it should be fixed into a separate PR.

timonv:
Nice! Currently all tree-sitter languages are exposed via the tree-sitter feature flag in `swiftide/swiftide`, could you add it there as well?

> Very cool! I suppose we don't have any tests for the refs/defs module.
Indeed, only a very basic one is present in the transformer. That is very odd, I recall writing them for each language 🤔 Would be greatly appreciated if at least a simple test for refs and defs is added there.

I'll fix the linting issue tomorrow.

timonv (changes requested):
Nice! Currently all tree-sitter languages are exposed via the tree-sitter feature flag in swiftide/swiftide, could you add it there as well?

> Very cool! I suppose we don't have any tests for the refs/defs module.
Indeed, only a very basic one is present in the transformer. That is very odd, I recall writing them for each language 🤔 Would be greatly appreciated if at least a simple test for refs and defs is added there.

I'll fix the linting issue tomorrow.

timonv:
> Nice! Currently all tree-sitter languages are exposed via the tree-sitter feature flag in swiftide/swiftide, could you add it there as well?

It's late, it's a single feature flag, my bad

timonv:
Master should be fixed! I've updated the branch

PR #336, opened 2024-09-25
feat(indexing)!: Removed duplication of batch_size. Pipeline owns the default ba…

### LINKED ISSUE #233: Single definition of batch size in indexing vs transformers
Batch size in indexing should either be on the pipeline only and inherited by the transformer, or set on the transformer only and used by the pipeline.

EmbeddingModels currently have a double batch size, it's confusing and not needed. I think the latter with a sane default could have the cleanest api.

devsprint:
Fixes #233

BREAKING CHANGE: The batch size of batch transformers when indexing is now configured on the batch transformer. If no batch size or default is configured, a configurable default is used from the pipeline. The default batch size is 256.

### DIFF
--- a/swiftide-indexing/src/pipeline.rs
+++ b/swiftide-indexing/src/pipeline.rs
@@ -11,6 +11,9 @@ use std::{sync::Arc, time::Duration};
 
 use swiftide_core::indexing::{EmbedMode, IndexingStream, Node};
 
+/// The default batch size for batch processing.
+const DEFAULT_BATCH_SIZE: usize = 256;
+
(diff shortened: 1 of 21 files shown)


swiftide-indexing/src/pipeline.rs
```diff
@@ -207,7 +212,6 @@ impl Pipeline {
     ///
```
timonv, line 212: A hint here that it will prefer the batch size from the transformer if it's present could save some headaches.

timonv (changes requested):
This is excellent, thank you (again) for the contribution! Just one small comment on docs.

timonv:
fyi #334 relates but it shouldn't interfere

PR #346, opened 2024-09-29
feat(indexing): Support Redb as embedable nodecache

timonv:
Adds support for Redb as an embeddable node cache, allowing full local app development without needing external services.


timonv:
image
Benchmark results

PR #378, opened 2024-10-11
fix: Add missing `Embed::batch_size` implementation

rcjmurillo:
`transformers::Embed` was missing the `Embed::batch_size` implementation for `BatchableTransformer`, causing the pipeline to always use the default batch size when using this transformer.


timonv:
Indeed, nice find! Thank you for your contribution

timonv (approved):

Issue #158, opened 2024-07-12
Postgres Storage


timonv:
Postgres with PGVector is popular for good reasons and it makes sense to support it. There are two parts to the integration, implementing `Persist` and implementing (various) search strategies.

Before implementation; since Postgres is not schemaless, I'm especially interested in a setup that alligns with other forms of storage and how we deal with additional fields (ie metadata).

shamb0:
Hello @timonv , I am interested to work on this, Could you please assign it to me.

timonv:
Hi @shamb0, thanks for your interest, ofcourse! I'd suggest to split up the PRs in (at least) two, indexing and querying seperately. What are your thoughts on schema management?

shamb0:
Thanks @timonv, Sounds Good. I will make two PR's. :-)

PR #392, opened 2024-10-16
feat(integrations): Add PGVector support for indexing

shamb0:
**Related to**: [#158](https://github.com/bosun-ai/swiftide/issues/158)

### PR Overview (Part-01)
This PR introduces the first part of the indexing pipeline, which handles the process of loading, chunking, enhancing, embedding, and storing markdown content into the `pgvector` store.

### How to Run the Example

The provided example, `swiftide/examples/index_md_into_pgvector.rs`, uses the Ollama LLM provider with the `llama3.2:latest` model for generating QA metadata.

**Steps to Run:**

1. **Start the PostgreSQL services**:
    ```bash
    docker compose -f ./scripts/docker/docker-compose-db-pg.yml up
    ```

2. **Run the example with the default configuration**:
    ```bash
    RUST_LOG=debug cargo run --example index-md-pgvector
    ```


swiftide-integrations/src/pgvector/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+    ///
+    /// # Returns
+    ///
+    /// * `Option<&PgPool>` - A reference to the connection pool if it exists, otherwise `None`.
+    ///
+    /// # Arguments
+    ///
+    /// None
+    ///
+    /// # Errors
```
timonv, line 96: There's no need to document arguments or errors if they cannot happen.

timonv (changes requested):
Nice job, this looks already very solid! I have a couple of comments and change requests. Notably the N+1 is important.

Additionally, I'm also missing tests. Could you add tests for Persist, batch and non batch? For lance and qdrant I used testcontainers, see theirs for an example, it works pretty well. They're in their retrieve implementation iirc.

For the example test dataset, could you remove that and use the root readme instead? For the docker compose yml, I'm not sure, either we should have one at the root, leave it out, or use testcontainers in examples as well.

Really nice job on this, it's looking good!

swiftide-integrations/src/pgvector/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+use derive_builder::Builder;
+use sqlx::postgres::PgPoolOptions;
+use sqlx::PgPool;
+use std::fmt;
+
+use pgv_table_types::{FieldConfig, MetadataConfig, VectorConfig};
+
+const PG_POOL_MAX_CONN: u32 = 10;
+const DEFAULT_BATCH_SIZE: usize = 50;
+const DEFAULT_VEC_DIM: i32 = 384;
```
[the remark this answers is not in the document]
shamb0: That makes sense! I’ve updated it similar to the LanceDB implementation, making `vector_size` a required configuration.

swiftide-integrations/src/pgvector/mod.rs
shamb0: Ah, I see. That change came from fixing some Clippy warnings. I’ve refactored the code accordingly. It should be fine now.

swiftide-integrations/src/pgvector/mod.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub async fn try_from_url(
+        mut self,
+        url: impl AsRef<str>,
+        connection_max: Option<u32>,
+    ) -> Result<Self> {
+        self.connection_pool = Some(Some(
+            PgPoolOptions::new()
+                .max_connections(connection_max.unwrap_or(PG_POOL_MAX_CONN))
+                .connect(url.as_ref())
+                .await?,
```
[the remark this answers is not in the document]
shamb0: Thanks for the suggestion! I’ve refactored the code accordingly, but please let me know if I’ve misunderstood anything.

I’ve wrapped `PgPool` inside the `PgDBConnectionPool` tuple struct and made `connection_pool` non-optional. The default is now handled via the builder macro with `PgDBConnectionPool::default()`, and the connection to the Postgres server is established through the `try_connect_to_pool()` method.

```rust
#[builder(default = "PgDBConnectionPool::default()")]
connection_pool: PgDBConnectionPool,
```

Let me know if you think further improvements are needed.

swiftide-integrations/src/pgvector/pgv_table_types.rs
```diff
@@ -0,0 +1,228 @@
...
+
+            let query = self.bind_node_to_query(&sql, node)?;
+
+            query.execute(&mut *tx).await.map_err(|e| {
+                tracing::error!("Failed to store node {}: {:?}", i, e);
+                anyhow!("Failed to store node {}: {:?}", i, e)
+            })?;
+
+            tracing::debug!("Successfully stored node {}", i);
+        }
```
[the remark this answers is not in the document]
shamb0: Good catch! I’ve refactored the code to perform batch inserts instead of executing `N` queries.

shamb0:
@timonv, Thanks a lot! I really appreciate your detailed feedback. I’ve addressed most of the points, but I’d like some clarification on your comment about the `PgVector::connection_pool` field. Could you confirm if I’ve understood it correctly, or if further changes are needed?

I’ve also added unit tests, following the primary use cases in LanceDB, and included tests for `retrieve`. Everything seems to be working fine so far (https://github.com/bosun-ai/swiftide/pull/406). The current PR is still in draft, and I plan to rebase it after the current intake.

examples/index_md_into_pgvector.rs
```diff
@@ -0,0 +1,77 @@
+/**
+* This example demonstrates how to use the Pgvector integration with SwiftIDE
```
timonv, line 2:
```suggestion
* This example demonstrates how to index markdown into PGVector
```

examples/index_md_into_pgvector.rs
```diff
@@ -0,0 +1,77 @@
...
+    let pgv_storage = PgVector::builder()
+        .try_connect_to_pool(pgv_db_url, Some(10))
+        .await
+        .expect("Failed to connect to postgres server")
+        .vector_size(384)
+        .with_vector(EmbeddedField::Combined)
+        .with_metadata(METADATA_QA_TEXT_NAME)
+        .table_name("swiftide_pgvector_test".to_string())
+        .build()
+        .unwrap();
```
timonv, line 52: Almost exactly right! I prefer it if builders do not do IO if they can avoid it, for multiple reasons. In this case, that also has the benefit of being able to connect lazilly and hiding the details of the connection pool. 

i.e. the builder api like:
```rust
 let pgv_storage = PgVector::builder()
        .database_url(pgv_db_url)
        .pool_size(10) // With a sane default if ommitted 
        .vector_size(384)
        .with_vector(EmbeddedField::Combined)
        .with_metadata(METADATA_QA_TEXT_NAME)
        .table_name("swiftide_pgvector_test".to_string())
        .build()
        .unwrap();
```

And then in `PgVector::setup` (which is only called once):

```rust
async fn setup(&self) -> Result<()> {
  self.try_connect_to_pool(self.database_url, self.pool_size).await?;
  ...
```

swiftide-integrations/src/pgvector/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+    #[builder(default = "PgDBConnectionPool::default()")]
+    connection_pool: PgDBConnectionPool,
+
+    /// Table name to store vectors in.
+    #[builder(default = "String::from(\"swiftide_pgv_store\")")]
+    table_name: String,
+
+    /// Default sizes of vectors. Vectors can also be of different
+    /// sizes by specifying the size in the vector configuration.
+    vector_size: Option<i32>,
```
timonv, line 33: This value isn't optional, is it? What happens if it is None?

swiftide-integrations/src/pgvector/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+
+    /// Batch size for storing nodes.
+    #[builder(default = "Some(DEFAULT_BATCH_SIZE)")]
+    batch_size: Option<usize>,
```
timonv, line 37: Is it intentional that this value is optional and can be None?

swiftide-integrations/src/pgvector/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+    ///
+    /// Supports multiple field types; see [`FieldConfig`] for details.
+    #[builder(default)]
+    fields: Vec<FieldConfig>,
+}
+
+impl fmt::Debug for PgVector {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        // Access the connection pool synchronously and determine the status.
+        let connection_status = self.connection_pool.connection_status();
```
timonv, line 49: Does this do a query? Debug is called extensively with `tracing`. Which would mean (I think?) a query on every log / trace statement.

swiftide-integrations/src/pgvector/mod.rs
```diff
@@ -0,0 +1,163 @@
...
+    /// This function returns an error if the connection to the database fails or if retries are exhausted.
+    /// Possible reasons include invalid database URLs, unreachable servers, or exceeded retry limits.
+    pub async fn try_connect_to_pool(
+        mut self,
+        url: impl AsRef<str>,
+        connection_max: Option<u32>,
+    ) -> Result<Self> {
+        let pool = self.connection_pool.clone().unwrap_or_default();
+
+        self.connection_pool = Some(pool.try_connect_to_url(url, connection_max).await?);
```
timonv, line 116: See other comment. Changing that and moving it to the main struct will also remove the need to clone, unwrap and option.

swiftide-integrations/src/pgvector/persist.rs
```diff
@@ -0,0 +1,124 @@
...
+    #[test_log::test(tokio::test)]
+    async fn test_persist_setup_no_error_when_table_exists() {
+        let test_context = TestContext::setup().await.expect("Test setup failed");
+
+        test_context
+            .pgv_storage
+            .setup()
+            .await
+            .expect("PgVector setup should not fail when the table already exists");
+    }
```
timonv, line 123: Could you add tests for storing nodes and verifying with a query that it got persisted as expected? For all the different embedding modes and with/without metadata.

timonv (changes requested):
Hey @shamb0,

Getting there! I've provided some more feedback. I still need to run the example to see how it performs, will do that this week.

swiftide-integrations/src/pgvector/mod.rs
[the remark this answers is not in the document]
timonv: Yes exactly, almost! I've left some comments

swiftide-integrations/src/pgvector/pgv_table_types.rs
```diff
@@ -0,0 +1,417 @@
...
+        let sql = self.generate_bulk_insert_sql(nodes.len());
+
+        let mut query = sqlx::query(&sql);
+
+        for node in nodes {
+            query = self.bind_node_to_bulk_query(query, node)?;
+        }
+
+        query.execute(&mut *tx).await.map_err(|e| {
+            tracing::error!("Failed to store nodes: {:?}", e);
```
timonv, line 260: It's good practice not to log errors in libraries if the error is bubbled up, as the caller is expected to handle it. Could you remove it?

swiftide-integrations/src/pgvector/pgv_table_types.rs
```diff
@@ -0,0 +1,417 @@
...
+    }
+
+    /// Generates an SQL upsert statement based on the current fields and table name.
+    ///
+    /// This function constructs a SQL query that inserts new rows into a `PostgreSQL` database table
+    /// if they do not already exist (based on the "id" column), or updates them if they do. The generated
+    /// SQL is intended to be efficient and safe for concurrent use.
+    #[allow(clippy::redundant_closure_for_method_calls)]
+    /// Generates a bulk insert SQL statement for inserting multiple nodes.
+    fn generate_bulk_insert_sql(&self, node_count: usize) -> String {
```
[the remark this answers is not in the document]
shamb0: - Thanks @timnov for suggesting `UNNEST`; that sounds promising! I'll give it a try and keep you posted.

- **Regarding the question, “What happens if the node is already present?”**
    - In the current setup, if a node already exists, a duplicate entry is created. I considered using a caching layer, like Redis, to manage duplicate source files after the `FileLoader` stage, rather than handling it at the storage level.

- **On the suggestion to “upsert on the ID of the node”**
    - Upserting on the node’s ID could lead to data loss. Here’s why: 

      Let’s take a look at the indexing pipeline structure:

      ```rust
      indexing::Pipeline::from_loader(FileLoader::new(test_dataset_path).with_extensions(&["md"]))
          .then_chunk(ChunkMarkdown::from_chunk_range(10..2048))
          .then(MetadataQAText::new(llm_client.clone()))
          .then_in_batch(Embed::new(fastembed.clone()).with_batch_size(100))
          .then_store_with(pgv_storage.clone())
          .run()
          .await?;
      ```

      - **Explanation**: In this example, the pipeline starts with `FileLoader`, which processes a single file and creates a node, let’s call it `N1`.
      - In the next stage, `ChunkMarkdown`, `N1` is split into multiple nodes (`one-to-X` relationship), each inheriting the same `ID` as `N1`.
      - These nodes, sharing the same ID, are then passed to the storage stage (`pgv_storage`).

    If we use upsert based on the ID, data from previously processed nodes may be overwritten, leading to possible data loss. I’d be glad to hear your thoughts on this or correct any misunderstandings if there’s a different intended approach.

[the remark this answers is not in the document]
timonv: It shouldn't be an issue, the id is lazilly created based on a hash of the chunk (usually at index time). Do you think there is a bug here? You might be on to something, I recently redid id generation such that it's lazy, but it indeed seems that it copies over the id when chunking, while it shouldn't. Perhaps the id shouldn't be lazy at all.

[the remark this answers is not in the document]
timonv: @shamb0 I think you identified a pretty hefty potential bug. Nodes should never share the same ID unless their chunk and path is exactly the same. As long as that holds true it should be safe to upsert. What could happen (maybe with a node cache?), that the id gets set earlier and gets copied over during chunking, never updating it again. That's not intentional and is a bug. I've fixed it in #414, thanks!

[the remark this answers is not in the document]
shamb0: Thanks, @timnov. I’m glad our discussion led to identifying this bug. Apologies for the delay—I’ve been managing some personal matters but am back on this task now. I’ll start working with patch #414 today and keep you updated on any progress.

examples/index_md_into_pgvector.rs
shamb0: @timonv, I'm looking for your input on a design choice here.

If we decide to handle the database connection pool setup within `fn setup(&self)` instead of `PgVectorBuilder`, we'll need to mutate `PgVector` within `fn setup()`. This change would mean updating the function signature in `trait Persist` to:

```rust
async fn setup(&mut self) -> Result<()>
```

For example:

```rust
async fn setup(&mut self) -> Result<()> {
  self.connection_pool = self.try_connect_to_pool(self.database_url, self.pool_size).await?;
  ...
}
```

This adjustment would introduce breaking changes across the stack, particularly impacting:

- `swiftide-indexing/src/persist/memory_storage.rs`
- `swiftide-integrations/src/lancedb/persist.rs`
- `swiftide-integrations/src/qdrant/persist.rs`
- `swiftide-integrations/src/redis/persist.rs`

Would you prefer moving the IO operations into `Persist::setup()` for these components? If so, we could handle this as a separate PR to streamline the updates.

Looking forward to your thoughts!

swiftide-integrations/src/pgvector/mod.rs
shamb0: For the `batch_size` configuration, I decided to keep it optional, drawing inspiration from the reference Qdrant implementation. By default, the batch size for PGVector is set to `50`—though we still plan to fine-tune this for optimal performance.

To answer your question, *"Is it intentional that this value is optional and can be None?"*—yes, it’s designed that way. In the indexing pipeline, `Pipeline::then_store_with()` functions as an adapter, routing `Node Streams` to backend storage depending on the batch setting.

Here’s the approach:
- When `Persist::batch_size()` returns `None`, each node is processed individually, with `Persist::store()` sending each chunk to the backend.
- If `Persist::batch_size()` returns `Some()`, batch processing is enabled. The stream of nodes is grouped into chunks based on the batch size, and `Persist::batch_store()` sends these batches to storage.

Does this implementation align with requirements? Let me know if there’s anything specific you’d like adjusted.

swiftide-integrations/src/pgvector/mod.rs
shamb0: Just to clarify: this parameter can’t be `None` because if it is, the user won’t be able to build the `PgVector::fields` parameter, and `PgVectorBuilder::with_vector()` would fail. We’re ensuring that users configure this parameter correctly before launching the indexing pipeline.

I’d love to hear your thoughts and any suggestions for enhancing this approach.

swiftide-test-utils/src/test_utils.rs
```diff
@@ -70,6 +73,41 @@ pub async fn start_redis() -> (ContainerAsync<GenericImage>, String) {
...
+/// Returns container server and `server_url`.
+pub async fn start_postgres() -> (ContainerAsync<GenericImage>, String, TempDir) {
+    // Create a temporary directory for Postgres data
+    let temp_dir = TempDir::new().expect("Failed to create temp folder");
+    let temp_data_bind_path = temp_dir.path().to_str().unwrap();
+
+    // Find a free port on the host for Postgres to use
+    let host_port = portpicker::pick_unused_port().expect("No available free port on the host");
+
+    let postgres = testcontainers::GenericImage::new("ankane/pgvector", "v0.5.1")
```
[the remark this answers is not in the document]
shamb0: Sounds good—I’ve migrated to the latest image, `pgvector/pgvector:pg17`:

```rust
let postgres = testcontainers::GenericImage::new("pgvector/pgvector", "pg17")
```

Let me know if there’s anything else you’d suggest here!

swiftide-test-utils/src/test_utils.rs
```diff
@@ -70,6 +73,41 @@ pub async fn start_redis() -> (ContainerAsync<GenericImage>, String) {
...
+            "database system is ready to accept connections",
+        ))
+        .with_mapped_port(host_port, 5432.tcp())
+        .with_env_var("POSTGRES_USER", "myuser")
+        .with_env_var("POSTGRES_PASSWORD", "mypassword")
+        .with_env_var("POSTGRES_DB", "mydatabase")
+        .with_mount(Mount::bind_mount(
+            temp_data_bind_path,
+            "/var/lib/postgresql/data",
+        ))
```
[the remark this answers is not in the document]
shamb0: Given the resource constraints in the CI/CD build context, `Mount::tmpfs_mount()` seems like a promising choice since the volume resource is released once the container is stopped.

```rust
let postgres = testcontainers::GenericImage::new("pgvector/pgvector", "pg17")
    .with_wait_for(WaitFor::message_on_stdout(
        "database system is ready to accept connections",
    ))
    .with_mapped_port(host_port, 5432.tcp())
    .with_env_var("POSTGRES_USER", "myuser")
    .with_env_var("POSTGRES_PASSWORD", "mypassword")
    .with_env_var("POSTGRES_DB", "mydatabase")
    .with_mount(Mount::tmpfs_mount("/var/lib/postgresql/data"))
    .start()
    .await
    .expect("Failed to start Postgres container");
```

Does this approach fit the context, or would you suggest any adjustments?

swiftide-integrations/src/pgvector/pgv_table_types.rs
shamb0: Sounds good! I've cleaned up the extra logs and traces.

shamb0:
- Thanks, @timnov, for the second round of review comments. I believe around 90% of the points have been addressed.
- The `Retrieve` functionality integration is complete.
- For unit test coverage, I've included critical scenarios and kept the scope focused to facilitate the review process. Let me know if you'd like to see any additional use cases covered—I’d appreciate your feedback.
- I’m currently troubleshooting a failure with `EmbedMode::PerField`. I’ll keep you posted on any progress there.

timonv:
> * Thanks, @timnov, for the second round of review comments. I believe around 90% of the points have been addressed.
> 
>     * The `Retrieve` functionality integration is complete.
> 
>     * For unit test coverage, I've included critical scenarios and kept the scope focused to facilitate the review process. Let me know if you'd like to see any additional use cases covered—I’d appreciate your feedback.
> 
>     * I’m currently troubleshooting a failure with `EmbedMode::PerField`. I’ll keep you posted on any progress there.

Perfect, let me know if I can help / continue the review

shamb0:
> > * Thanks, @timnov, for the second round of review comments. I believe around 90% of the points have been addressed.
> >   
> >   * The `Retrieve` functionality integration is complete.
> >   * For unit test coverage, I've included critical scenarios and kept the scope focused to facilitate the review process. Let me know if you'd like to see any additional use cases covered—I’d appreciate your feedback.
> >   * I’m currently troubleshooting a failure with `EmbedMode::PerField`. I’ll keep you posted on any progress there.
> 
> Perfect, let me know if I can help / continue the review

Yes @timonv, I need your help regarding the discussion thread on (https://github.com/bosun-ai/swiftide/pull/392#discussion_r1822229042). 

Could you review my recent comments and share any updated thoughts when convenient? Once I have your input, I can proceed with the implementation, and then we can move forward to the intake review process.

examples/index_md_into_pgvector.rs
timonv: Hmm good point. Since it runs only once and only during setup, perhaps inner mutability solves it nicer. I.e. with https://docs.rs/tokio/latest/tokio/sync/struct.OnceCell.html

Then:
```rust
async fn setup(&self) -> Result<()> {
  self.connection_pool.set(self.try_connect_to_pool().await?).await?;
  ...
}

// Where
struct PGVector {
  connection_pool: OnceCell<PGPool>
}

// And have a small wrapper to avoid the double error error handling like:
fn get_pool(&self) -> Result<&PGPool>
```

swiftide-integrations/src/pgvector/mod.rs
timonv: Why make it an Option then? On the builder side it's still an option, when build, It shouldn't be.

timonv:
Really nice job so far! It's a lot of work. It looks like you accidently merged in the Retrieve and query stuff, could you split it out again?

examples/index_md_into_pgvector.rs
shamb0: @timonv, Thanks for the suggestion! It looks promising it should work well. I'll make the adjustments and update you shortly.

shamb0: Thank you for the valuagble suggestion, @timnov! I've completed the refactoring, and I believe the implementation is now cleaner and more efficient. 

Here’s a snippet of the updated code:

```rust
pub struct PgVector {
    ...

    /// Lazy-initialized database connection pool.
    #[builder(default = "Arc::new(OnceCell::new())")]
    connection_pool: Arc<OnceCell<PgPool>>,
}
```

Thanks again for the input!

swiftide-integrations/src/pgvector/mod.rs
shamb0: Thank you for this review feedback!

You made an excellent point about the `vector_size` and `batch_size` fields not requiring the `Option<T>` type. I've updated the PgVector implementation accordingly:

```rust
pub struct PgVector {
    /// Default vector size; can be customized per configuration.
    vector_size: i32,

    /// Batch size for storing nodes.
    #[builder(default = "BATCH_SIZE")]
    batch_size: usize,
}
```

I missed that detail when working on the `PgVectorBuilder` build function. I've completed the refactoring and cleanup.

shamb0:
Hi @timnov,

Thank you for your input! I believe the patch is ready for the final intake review.

- Removed the retrieval functionality.
- Addressed all intake blockers, with refactoring and cleanup completed.
- Verified the persistence functionality across all `EmbedMode` combinations, and everything is working smoothly with the unit tests.

Please let me know if any further enhancements are needed.

shamb0:
Hi @timonv,

I wanted to bring to your attention some CI/CD build test failures I've noticed. When running `cargo test --all-features --tests`, I'm seeing consistent failures, and interestingly, these are also present in the `upstream/master` branch.

I'd appreciate your thoughts on this and would be happy to help address these issues if needed.

timonv:
> Hi @timonv,
> 
> I wanted to bring to your attention some CI/CD build test failures I've noticed. When running `cargo test --all-features --tests`, I'm seeing consistent failures, and interestingly, these are also present in the `upstream/master` branch.
> 
> I'd appreciate your thoughts on this and would be happy to help address these issues if needed.

Thanks, it should be a lot better since #434. With all the integrations it's a bit of an uphill battle trying to not go over free github action limits

swiftide-integrations/src/pgvector/pgv_table_types.rs
```diff
@@ -0,0 +1,481 @@
...
+            INSERT INTO {} ({})
+            SELECT {}
+            FROM UNNEST({}) AS t({})
+            ON CONFLICT (id) DO UPDATE SET {}"#,
+            self.table_name,
+            columns.join(", "),
+            columns.join(", "),
+            unnest_params.join(", "),
+            columns.join(", "),
+            update_columns
```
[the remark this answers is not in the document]
shamb0: Hi @timonv,

I apologize for the delay.

- Regarding your comment "Additionally, it's generated for each bulk insert, but I think it's identical for the whole pipeline, right?"

Yes you were right that the SQL statement for bulk insert is applied to the entire pipeline. Optimizing this could be achieved by introducing a new state variable in `PgVector` and initializing it as part of `setup()`, does it makesense? :

```rust
#[derive(Builder, Clone)]
#[builder(setter(into, strip_option), build_fn(error = "anyhow::Error"))]
pub struct PgVector {
    sql_stmt_bulk_insert: String,
}
```

- For your comment "If columns or table names include sql, this is vulnerable to sql injection. There has to be a way to avoid this?", I need your help and more details. Please find the SQL statement used for bulk insert:

```sql
INSERT INTO swiftide_pgvector_test (
    id,
    chunk,
    vector_combined,
    meta_questions_and_answers
)
SELECT 
    id,
    chunk,
    vector_combined,
    meta_questions_and_answers
FROM UNNEST(
    $1::UUID[],
    $2::TEXT[],
    $3::VECTOR[],
    $4::JSONB[]
) AS t(
    id,
    chunk,
    vector_combined,
    meta_questions_and_answers
)
ON CONFLICT (id) DO UPDATE 
SET 
    chunk = EXCLUDED.chunk,
    vector_combined = EXCLUDED.vector_combined,
    meta_questions_and_answers = EXCLUDED.meta_questions_and_answers;
```

Could you please clarify what might be causing the SQL injection vulnerability? Are there any issues with the naming conventions used for `table` and `column_name`, or is it related to the usage of placeholders?

Thank you for your help.

Cargo.toml
```diff
@@ -40,6 +39,7 @@ indoc = { version = "2.0" }
 regex = { version = "1.11.1" }
 uuid = { version = "1.10", features = ["v3", "v4", "serde"] }
 dyn-clone = { version = "1.0" }
+once_cell = { version = "1.20.2" }
```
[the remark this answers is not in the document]
shamb0: - I initially chose `once_cell` because `std::cell::OnceCell` is not thread-safe, which resulted in the following error.  
- I have now switched to `std::sync::OnceLock`, which is thread-safe and part of the standard library, thereby eliminating the dependency on the `once_cell` crate.

```shell
error[E0277]: `std::cell::OnceCell<sqlx::Pool<sqlx::Postgres>>` cannot be shared between threads safely
   --> swiftide-integrations/src/pgvector/persist.rs:13:18
    |
13  | impl Persist for PgVector {
    |                  ^^^^^^^^ `std::cell::OnceCell<sqlx::Pool<sqlx::Postgres>>` cannot be shared between threads safely
    |
    = help: the trait `std::marker::Sync` is not implemented for `std::cell::OnceCell<sqlx::Pool<sqlx::Postgres>>`, which is required by `pgvector::PgVector: std::marker::Sync`
```

Cargo.toml
```diff
@@ -1,6 +1,5 @@
 [workspace]
 members = ["swiftide", "swiftide-*", "examples", "benchmarks"]
-default-members = ["swiftide", "swiftide-*"]
```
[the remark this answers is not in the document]
shamb0: I apologize for the oversight that occurred during the re-basing of the workspace.

swiftide-integrations/src/pgvector/pgv_table_types.rs
```diff
@@ -0,0 +1,481 @@
...
+                            anyhow!("Metadata field {} not found", config.original_field)
+                        })?;
+
+                        let entry = bulk_data
+                            .metadata_fields
+                            .entry(config.field.clone())
+                            .or_default();
+
+                        let mut metadata_map = BTreeMap::new();
+                        metadata_map.insert(config.original_field.clone(), value.clone());
```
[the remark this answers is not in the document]
timonv: Nice solution!

timonv (dismissed):
Awesome! Really nice job and thank you for bearing with me, it's a big and complicated feature!

I'll take a look at the pipelines, and then merge it asap. I'd love to mention you on linkedin and other socials, feel free to connect (@timonv everywhere).

🚀

shamb0:
> Awesome! Really nice job and thank you for bearing with me, it's a big and complicated feature!
> 
> I'll take a look at the pipelines, and then merge it asap. I'd love to mention you on linkedin and other socials, feel free to connect (@timonv everywhere).
> 
> 🚀

Thanks @timonv!

Really appreciate your thorough guidance throughout this feature implementation. Your insight on the runtime performance issue was particularly valuable - it helped prevent potential pipeline bottlenecks before they could become problems.

There's one remaining discussion point that needs attention when you have a moment:
https://github.com/bosun-ai/swiftide/pull/392#discussion_r1847640953

And thanks for the social media invitation - I'll definitely connect with you (@timonv)!

Cheers

swiftide-integrations/src/pgvector/pgv_table_types.rs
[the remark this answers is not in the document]
timonv: As for optimizing the bulk insert, a local static perhaps. For the sql injection I suspect the column names might be vulnerable as they're generated at runtime without escaping, but the combination of it producing a working query and it actually happening (table names from user/ai input?) are minor. I'm fine with leaving it as is, can always be picked up later.

timonv:
Tuned feature flags and the ci a bit, let's see where it goes. #450 might also have some impact and includes a fix for the deny.toml, I'll merge that in a bit.

PR #463, opened 2024-11-27
feat(agents): **EXPERIMENTAL** Agents in Swiftide

timonv:
Agents are coming to Swiftide! We are still ironing out all the kinks, while we make it ready for a proper release. You can already experiment with agents, see the rustdocs for documentation, and an example in `/examples`, and feel free to contact us via github or discord. Better documentation, examples, and tutorials are coming soon.

Run completions in a loop, define tools with two handy macros, customize the agent by hooking in on lifecycle events, and much more.

Besides documentation, expect a big release for what we build this for soon! 🎉


swiftide-core/src/agent_traits.rs
```diff
@@ -0,0 +1,107 @@
...
+
+impl From<String> for CommandOutput {
+    fn from(value: String) -> Self {
+        CommandOutput::Text(value)
+    }
+}
+
+/// Acts as the interface to the external world and manages messages for completion
+#[async_trait]
+pub trait AgentContext: Send + Sync {
```
tinco, line 79: This trait feels too large. I think at the very least it should be split up into a trait for each concern, and then have it include those traits. I'd perhaps give it a slightly more unique too.

i.e. it's an async messaging context, and a history keeping context, and a context that allows command execution. Since it's so opinionated, maybe it should be called a SwiftideAgent or maybe even more specific to allow for later iterations.

Issue #511, opened 2024-12-30
Implementation of `CustomStrategy` for LanceDB and Qdrant


timonv:
**Is your feature request related to a problem? Please describe.**
@shamb0 recently added a custom search strategy that makes it possible for users to generate a custom query, using the providers query builder, to easily customize queries for retrievers.

Still needs to be implemented for other providers, lancedb and qdrant.

**Describe the solution you'd like**
Qdrant has a straightforward query builder. Lance however has a query builder that is bound to the table. A `Retrieve` implementation similar to PGVector for the strategy would be the way to go.

See https://github.com/bosun-ai/swiftide/blob/master/swiftide-core/src/search_strategies/custom_strategy.rs

PR #518, opened 2024-12-31
feat(query): Add custom lancedb query generation for lancedb search

shamb0:
Fixes: #511

This PR implements the `CustomStrategy<Q>` trait for LanceDB, enabling flexible and customizable vector similarity search capabilities. The implementation follows the successful pattern established in the pgvector integration while leveraging LanceDB-specific features.

### DIFF
--- a/swiftide-integrations/src/lancedb/retrieve.rs
+++ b/swiftide-integrations/src/lancedb/retrieve.rs
@@ -1,6 +1,5 @@
 use anyhow::Result;
-use arrow::datatypes::SchemaRef;
-use arrow_array::StringArray;
+use arrow_array::{RecordBatch, StringArray};
 use async_trait::async_trait;
 use futures_util::TryStreamExt;
 use itertools::Itertools;
@@ -8,7 +7,10 @@ use lancedb::query::{ExecutableQuery, QueryBase};
 use swiftide_core::{
     document::Document,
     indexing::Metadata,
-    querying::{search_strategies::SimilaritySingleEmbedding, states, Query},
+    querying::{
+        search_strategies::{CustomStrategy, SimilaritySingleEmbedding},
+        states, Query,
+    },
     Retrieve,
 };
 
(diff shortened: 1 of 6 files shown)


swiftide/tests/lancedb.rs
```diff
@@ -87,3 +92,125 @@ async fn test_lancedb() {
...
+    let db_connection = lancedb
+        .get_connection()
+        .await
+        .expect("Failed to establish connection to LanceDB pool");
+
+    let vector_table = db_connection
+        .open_table("swiftide_test")
+        .execute()
+        .await
+        .expect("Failed to open vector search table 'swiftide_test'");
```
timonv, line 150: I wonder if this would keep the connection open, and force a single connection to be used for the full query pipeline. That might be an unintentional bottleneck.

swiftide/tests/lancedb.rs
```diff
@@ -87,3 +92,125 @@ async fn test_lancedb() {
...
+
+    let similarity_config =
+        swift_query_pipeline::search_strategies::SimilaritySingleEmbedding::from_filter(
+            "filter = \"true\"".to_string(),
+        );
+
+    // Create the custom query strategy for vector similarity search
+    let vector_search_strategy =
+        swift_query_pipeline::search_strategies::CustomStrategy::from_query(
+            move |query_node| -> Result<lance_query_builder::VectorQuery> {
```
timonv, line 160: If that's the case, maybe (thinking very meta), it's possible to yield the strategy here too. And then having the concrete implementation of the strategy (so specific for lance), yield an active table connection.

Honestly, it sounds rather complicated, I don't have an off the bat solution

swiftide-integrations/src/lancedb/retrieve.rs
```diff
@@ -124,6 +127,69 @@ impl Retrieve<SimilaritySingleEmbedding> for LanceDB {
     }
 }
 
+#[async_trait]
+impl Retrieve<CustomStrategy<query::VectorQuery>> for LanceDB {
```
timonv, line 131: Lancedb has a (not judging) bit confusing setup with concrete types starting at Query, which can be converted to a VectorQuery, they implement QueryBase and ExecutableQuery. I think ExecutableQuery is probably the type you want, but it's not dyn compatible (might still work with generics). Then a single implementation would work for both query types.

swiftide-integrations/src/lancedb/retrieve.rs
```diff
@@ -124,6 +127,69 @@ impl Retrieve<SimilaritySingleEmbedding> for LanceDB {
...
+                        } else {
+                            // Assume other fields are part of the metadata
+                            let value = array.value(row_idx).to_string();
+                            metadata.insert(field.name().clone(), value);
+                        }
+                    } else {
+                        // Handle other array types as necessary
+                        // TODO: Can't we just downcast to serde::Value or fail?
+                    }
+                }
```
timonv, line 176: Perhaps nice to refactor this to a `from_record_batch(batch) -> Documents`, so the code isn't duplicated across retriever implementations

swiftide/tests/lancedb.rs
```diff
@@ -87,3 +92,125 @@ async fn test_lancedb() {
...
+                let vector_field =
+                    lance_integration::VectorConfig::from(EmbeddedField::Combined).field_name();
+
+                let mut query_builder = vector_table
+                    .query()
+                    .nearest_to(embedding.as_slice())
+                    .context("Failed to create vector similarity query")?
+                    .column(&vector_field);
+
+                let result_limit = usize::try_from(similarity_config.top_k())
```
timonv, line 175: Maybe just use an arbitrary number? Feels a bit complicated to just create a new search strategy, if you really need a number.

timonv (changes requested):
Nice! Couple of comments

shamb0:
Hi @timonv,

I've updated the custom strategy implementation in `swiftide-core/src/search_strategies/custom_strategy.rs` to handle async query builders, addressing your feedback about LanceDB connection management. The documentation has been streamlined as well.

    https://github.com/bosun-ai/swiftide/pull/518#discussion_r1900417336
    https://github.com/bosun-ai/swiftide/pull/518#discussion_r1900417831

Would appreciate your thoughts on whether this prototype aligns with your expectations.

timonv (approved):
Nice work, this is really good! Will merge it asap.

PR #573, opened 2025-01-24
fix(tracing): Use `or_current()` to prevent orphaned tracing spans

tinco:
When a span is emitted that would be selected by the subscriber, but we instrument its closure with a span that would not be selected by the subscriber, the span would be emitted as an orphan (with a new `trace_id`) making them hard to find and cluttering dashboards.

This situation is also documented here: https://docs.rs/tracing/latest/tracing/struct.Span.html#method.or_current


timonv:
Nice! Should we also change the trace_span to info_spans? I don't have a usecase where I actually wouldn't want them.

timonv (approved):

PR #596, opened 2025-02-06
feat: add generics to allow azure integration

JoranHonig:
async-openai already supports connecting with the azure api, the only change necessary was adding generics.

This is technically a breaking change, however I've make the regular openai config the default limiting the downstream need for changes.

### DIFF
--- a/swiftide-integrations/src/openai/mod.rs
+++ b/swiftide-integrations/src/openai/mod.rs
@@ -10,8 +10,8 @@ mod embed;
 mod simple_prompt;
 
 // expose type aliases to simplify downstream use of the open ai builder invocations
-pub type OpenAIConfig = async_openai::config::OpenAIConfig;
-pub type AzureConfig = async_openai::config::AzureConfig;
+pub use async_openai::config::AzureConfig;
+pub use async_openai::config::OpenAIConfig;
 
 /// The `OpenAI` struct encapsulates an `OpenAI` client and default options for embedding and prompt models.
 /// It uses the `Builder` pattern for flexible and customizable instantiation.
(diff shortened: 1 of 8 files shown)


timonv (commented):
Thank you for the contribution(s), this looks very nice! It's also a nice step towards removing all the async_openai duplication (groq, openrouter, dashscope, ollama). Happy to merge it when the tests pass!

swiftide-integrations/src/openai/mod.rs
```diff
@@ -9,6 +9,10 @@ mod chat_completion;
 mod embed;
 mod simple_prompt;
 
+// expose type aliases to simplify downstream use of the open ai builder invocations
+pub type OpenAIConfig = async_openai::config::OpenAIConfig;
+pub type AzureConfig = async_openai::config::AzureConfig;
```
timonv, line 14: You should be able to do `pub use async_openai::config::{OpenAIConfig,AzureConfig}`, then rust/docs will also recognize it as a re-export

timonv:
Ok, nice work! The examples (they really should be in the default suite again -.-) fail because of missing generics. If it's ok with you, I'd like to make the following changes:

Have one generic openai-like implementation, i.e. `openai::GenericClient<C>`

Then add some type aliases, maintaining current export paths like:

`ollama::Client = openai::GenericClient<OllamaConfig>`
`azure::Client = openai::GenericClient<AzureConfig>`

etc.

I understand it's a bit out of scope of this contribution. For me it's not much work, I can pick it up within a couple of days. And it avoids (hopefully) any breaking changes.

timonv:
Fixed it, thank you for the contribution! I'll be releasing it asap

PR #597, opened 2025-02-06
feat: add solidity support

JoranHonig:
This pr adds solidity support. 

Notes:
* I've added queries, but have omitted outlining for now.
* The simple testcase should ensure the queries are correct ( I copied them from upstream anyways 🤷‍♂️ )

I had a quick question while writing this pr, is there any reason why the tags.scm file packaged with the parsers isn't used? or was this just easier for now.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -84,6 +84,7 @@ tree-sitter-ruby = "0.23"
 tree-sitter-rust = "0.23"
 tree-sitter-typescript = "0.23"
 tree-sitter-go = "0.23"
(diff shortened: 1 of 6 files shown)


timonv:
Cool! Looks good to me.

> I had a quick question while writing this pr, is there any reason why the tags.scm file packaged with the parsers isn't used? or was this just easier for now.

So far I had to change the queries slightly, they always included more captures than I needed.

As for the outlining, that's perfectly fine, with context sizes getting so big there's far less of a need

timonv:
Thank you, merging it in

PR #607, opened 2025-02-11
fix: make builder() generic

JoranHonig:
I noticed I forgot to make this generic 😅


timonv:
I actually made it concrete before merging, as otherwise it would have breaking changes (you would need to specify the generic parameter everytime you called OpenAi::builder). The builder itself should be public too, so you could use it directly.

PR #628, opened 2025-02-20
feat: Continue Agent on Tool Failure

bogzbonny:
Continues the agent on tool failure, related to https://github.com/bosun-ai/kwaak/issues/313

### DIFF
--- a/swiftide-agents/src/agent.rs
+++ b/swiftide-agents/src/agent.rs
@@ -103,10 +103,6 @@ pub struct Agent {
     /// the name and args of the tool.
     #[builder(private, default)]
     pub(crate) tool_retries_counter: HashMap<u64, usize>,
-
-    /// If true, the agent will continue to run even if a tool call fails after all retries
-    #[builder(default = false)]
-    pub(crate) continue_on_tool_total_failure: bool,
 }
 
 impl std::fmt::Debug for Agent {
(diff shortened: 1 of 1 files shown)


timonv:
That's interesting, let's try this out. Could you add/expand the limit test? I think it's the one at the bottom

timonv (changes requested):
See comment

timonv:
Do you think it shouldn't stop? My reasoning was that something is wrong, so put it back to the user instead. You're right that it needs a message to keep it balanced.

bogzbonny:
I'm actually not totally sure. Maybe it should stop? I suppose it's a kind of a "panic" condition... however I'm also thinking though that maybe there are situations which you _would_ want the agent to continue, maybe in a multi-agent model the parent agent would want to test out functionality of simpler agents which sometimes just wont work well (but maybe are cheap and cost effective) - in this situation you would want the agent to continue and maybe return to the parent agent noting that it errored out.

I suppose I'm kind of still airing on the side that we _should_ allow for continuation of the agent as an option (but maybe defaulted to OFF for now)

nomatter what though we need to keep the message balance!

What do you think @timonv?

timonv:
Maybe we're overthinking it. If tools continuously fail, I think the default behavior to stop is good enough (we could optionally have the limit disabled). I think the users problem is that you can't recover from the issue as the call/output balance is botched. So perhaps, a very simple solution is to just always push the message, even if the tool failed. Then, if the limit is reached, the agent is stopped, otherwise it just continuous as normal.

Practically I think that simplifies the code a lot as well. i.e.

```
log_error
push message
if limit reached
  agent.stop
```

I'm not sure how many people are using agents outside kwaak itself though. Maybe finer grained control is actually really useful for some cases.

bogzbonny:
Okay, based on both of our thinking, I think I'll remove the continuation portion of this PR... if we need it later we can add it then

PR #630, opened 2025-02-20
feat: Introduce LanguageModelError for LLM traits and an optional backoff decorator

tinco:
When there are unreliable network conditions, or when the language model service is unreliable, some Swiftide operations fail or drop results.

To ensure Swiftide never fails or drops results, especially on large datasets or long benchmarks it is therefore desirable that we can retry language model requests.

To retry requests however, we need to be sure if the error returned by the language model is due transient or permanent. We could use `backoff`'s `Error` enum for this, however there is one additional concern specific to language models and that is a specific kind of client error, the `ContextLengthExceeded` error. Even though this is a client error, it is desirable that the application is able to deal with this error at runtime. So we introduce the following type:

```rust
pub enum LanguageModelError {
    ContextLengthExceeded(BoxedError),
    TransientError(BoxedError),
    PermanentError(BoxedError),
}
```

The first is when the caller makes a mistake and sends too many tokens to the model. This might be recoverable, as the client could adjust its token amount and try again.

Then the second type of error is when there is any sort of network problem, server problem or other intermittent issue. In this case the prompt can be retried as-is and should eventually succeed.

The third type is when there is an error that can not be recovered from at runtime. The client made an error and the server will always respond with an error.

In addition to this, I have added a struct called `ReliableLanguageModel` that users can wrap around a language model to make it automatically recover from network/server issues.

I've added an example of its use to the `index-codebase` example.


swiftide-core/src/indexing_traits.rs
```diff
@@ -547,14 +588,79 @@ pub trait SimplePrompt: Debug + Send + Sync + DynClone {
...
+            .with_initial_interval(Duration::from_secs(self.config.initial_interval_sec))
+            .with_multiplier(self.config.multiplier)
+            .with_max_elapsed_time(Some(Duration::from_secs(self.config.max_elapsed_time_sec)))
+            .with_randomization_factor(self.config.randomization_factor)
+            .build()
+    }
+}
+
+#[async_trait]
+impl<P: SimplePrompt + Clone> SimplePrompt for ReliableLanguageModel<P> {
```
timonv, line 616: I'm a bit stuck on the name here. I think if it wraps the client with backoff, we should call it that. `Reliable` feels a bit like jibe / wishful thinking.

swiftide-integrations/src/anthropic/chat_completion.rs
```diff
@@ -58,7 +58,7 @@ impl ChatCompletion for Anthropic {
 
         let request = anthropic_request
             .build()
-            .map_err(|e| ChatCompletionError::LLM(Box::new(e)))?;
+            .map_err(|e| LanguageModelError::PermanentError(e.into()))?;
```
timonv, line 61: Small nit, these lines feel more verbose than they should, doesn't `.map_err(LanguageModelError::PermanentError)` work?

swiftide-integrations/src/anthropic/simple_prompt.rs
```diff
@@ -1,41 +1,72 @@
...
+        let response = self.client.messages().create(request).await.map_err(|e| {
+            let CreateMessagesError::AnthropicError(e) = e;
+            match e {
+                AnthropicError::NetworkError(_) => LanguageModelError::TransientError(e.into()),
+                AnthropicError::UnexpectedError => {
+                    LanguageModelError::PermanentError("Anthropic unexpected error".into())
+                    // Seriously?
+                }
+                AnthropicError::Unauthorized => {
+                    LanguageModelError::PermanentError("Anthropic unauthorized".into())
```
timonv, line 41: Also on 37, this hides the underlying error. Wouldn't it be better to always wrap the underlying error?

swiftide-integrations/src/anthropic/simple_prompt.rs
```diff
@@ -1,41 +1,72 @@
...
+                }
+                // TODO: The Rust Anthropic client is not documented well, we should figure out
+                // which of these errors are client errors and which are server errors.
+                // And which would be the ContextLengthExceeded error
+                // For now, we'll just map all of them to client errors so we get feedback.
```
timonv, line 46: I mean, I wrote it lol

swiftide-integrations/src/openai/mod.rs
```diff
@@ -211,6 +213,66 @@ impl<C: async_openai::config::Config + Default> GenericOpenAI<C> {
     }
 }
 
+pub fn openai_error_to_language_model_error(e: OpenAIError) -> LanguageModelError {
```
timonv, line 216: I suspect that if you replace this with a `impl From<OpenAIError> for LanguageModelError`, you can remove every single line with `.map_err(openai_error_to_language_model_error)` as along as there is a `?` at the end.

swiftide-integrations/src/openai/mod.rs
```diff
@@ -211,6 +213,66 @@ impl<C: async_openai::config::Config + Default> GenericOpenAI<C> {
...
+            } else {
+                // making the request failed for some other reason, probably recoverable
+                tracing::error!("Unexpected OpenAI Reqwest Error: {:?}", e);
+                LanguageModelError::TransientError(e.into())
+            }
+        }
+        OpenAIError::JSONDeserialize(e) => {
+            // OpenAI generated a non-json response, probably a temporary problem on their side
+            tracing::error!("OpenAI response could not be deserialized: {:?}", e);
+            LanguageModelError::TransientError(e.into())
```
timonv, line 255: This would be a cloudfare response then?

swiftide-integrations/src/openai/mod.rs
```diff
@@ -211,6 +213,66 @@ impl<C: async_openai::config::Config + Default> GenericOpenAI<C> {
...
+            LanguageModelError::PermanentError(OpenAIError::FileReadError(msg).into())
+        }
+        OpenAIError::StreamError(msg) => {
+            tracing::error!("OpenAI Stream failed: {:?}", msg);
+            LanguageModelError::PermanentError(OpenAIError::StreamError(msg).into())
+        }
+        OpenAIError::InvalidArgument(msg) => {
+            tracing::error!("OpenAI Invalid Argument: {:?}", msg);
+            LanguageModelError::PermanentError(OpenAIError::InvalidArgument(msg).into())
+        }
```
timonv, line 272: Are you sure you want to log _and_ return the error? Usually you only log errors if the error is handled or otherwise not bubbled up.

swiftide-integrations/src/openai/mod.rs
timonv: There's also some other places where there's map_err's but the type of error is always permanent (i.e. io errors). Potential to remove a lot of code everywhere.

timonv (commented):
Ok nice, a decorator so sounds like a nice idea so we can try this out. Do you also want to implement it for chat completions?

swiftide-integrations/src/anthropic/chat_completion.rs
tinco: oh shit yeah that would be a lot better

swiftide-integrations/src/anthropic/simple_prompt.rs
tinco: haha oh no, I forgot that you did, I'll check it out

swiftide-integrations/src/openai/mod.rs
tinco: Yeah most likely

swiftide-core/src/indexing_traits.rs
tinco: You're right, I also confused reliability with resilience, would `ResilientLanguageModel` be better? Or `LanguageModelWithBackoff`?

swiftide-integrations/src/openai/mod.rs
```diff
@@ -308,4 +309,96 @@ mod test {
...
+    #[test]
+    fn test_file_save_error_is_permanent() {
+        // Create a file save error
+        let openai_error = OpenAIError::FileSaveError("Failed to save file".to_string());
+        let result = openai_error_to_language_model_error(openai_error);
+
+        // Verify it's categorized as PermanentError
+        match result {
+            LanguageModelError::PermanentError(_) => {} // Expected
+            _ => panic!("Expected PermanentError, got {:?}", result),
```
timonv, line 362:
```rust
assert_matches!(result, LanguageModelError::PermanentError(..))
```

https://doc.rust-lang.org/std/assert_matches/macro.assert_matches.html

swiftide-integrations/src/openai/mod.rs
tinco: I can't `impl From<OpenAIError> for LanguageModelError` without adding `async_openai` as an (optional) dependency for swiftide core, or do you see another way?

swiftide-integrations/src/openai/mod.rs
tinco: nice maar dat is een nightly feature blijkbaar

tinco:
wtf, the ci / lint step suddenly started failing on files that aren't modified in this PR. Maybe a new version of clippy was released?

swiftide-integrations/src/openai/mod.rs
timonv: Huh yeah, really don't want that... Maybe  make the error associated in the trait and put it in integrations, or use an owned type pattern so you can do the impl in integrations, or approach the problem differently. Probably more ways. I don't know if it's worth going through all those hoops just to try this out. It would remove a lot of code though, so there's a clear upside there, up to you 🤷

swiftide-integrations/src/openai/mod.rs
timonv: Oh oops, you can still use `matches!` though, the above syntax shouldn't be necessary ever.

swiftide-core/src/indexing_traits.rs
timonv: Yeah I think the latter works. Can you think of. a better place to put it than in traits?

swiftide-integrations/src/openai/mod.rs
```diff
@@ -211,9 +213,60 @@ impl<C: async_openai::config::Config + Default> GenericOpenAI<C> {
...
+                // PermanentError depending on the message, if it contains the word
+                // quota, it should be a PermanentError otherwise it should
+                // be a TransientError.
+                // If the response code is any other 4xx it should be a PermanentError.
+                if status.as_u16() == 429 && !e.to_string().contains("quota") {
+                    LanguageModelError::TransientError(e.into())
+                } else if status.is_client_error() {
+                    LanguageModelError::PermanentError(e.into())
+                } else if status.is_server_error() {
+                    LanguageModelError::TransientError(e.into())
```
timonv, line 238: I'm not sure. What point is there in retrying if the remote server doesn't work? They had an outage this weekend, it would only have clobbered more resources on both ends than result in anything productive.

swiftide-integrations/src/openai/mod.rs
```diff
@@ -211,9 +213,60 @@ impl<C: async_openai::config::Config + Default> GenericOpenAI<C> {
...
+            }
+        }
+        OpenAIError::Reqwest(e) => {
+            if let Some(status) = e.status() {
+                // If the response code is 429 it could either be a TransientError or a
+                // PermanentError depending on the message, if it contains the word
+                // quota, it should be a PermanentError otherwise it should
+                // be a TransientError.
+                // If the response code is any other 4xx it should be a PermanentError.
+                if status.as_u16() == 429 && !e.to_string().contains("quota") {
```
timonv, line 233: I'm pretty sure the quote error is an ApiError, not a Reqwest error (https://github.com/64bit/async-openai/blob/c9acfed4d7434a5c3167f9ad95441bae0318badd/async-openai/src/client.rs#L357).

swiftide-integrations/src/openai/mod.rs
tinco: That's what backoff is for. If the servers are clobbered, it backs off exponentially until the servers are no longer clobbered. Additionally, if there is a more long term problem, then backoff eventually gives up too.

swiftide-integrations/src/openai/mod.rs
tinco: Yeah you're right. Unfortunately this also means we can not distinguish between server and client errors. Luckily though openai has now documented how they expect clients to deal with server errors (in our favor) so I've submitted a PR to async_openai to incorporate this logic https://github.com/64bit/async-openai/pull/354.

I've modified the code to rely on this behavior, though that means it won't work correctly until the PR is merged and a new version is released.

timonv (approved):
Nice work! I want to quickly test it with a longer running pipeline in kwaak before merging, otherwise looks great.

Cargo.toml
```diff
@@ -65,8 +65,10 @@ async-openai = { version = "0.28", default-features = false }
 qdrant-client = { version = "1.13", default-features = false, features = [
   "serde",
 ] }
-fluvio = { version = "0.26", default-features = false }
-lancedb = { version = "0.18", default-features = false }
+fluvio = { version = "0.26.2", default-features = false }
+lancedb = { version = "0.18", default-features = false, features = [
+  "remote",
+] }
```
timonv, line 71: Why was this feature added? I don't think it should be here

timonv (changes requested):
Overlooked the lancedb feature flag. I'll fix fluvio in master.

Cargo.toml
timonv: Ah just checked, that's unfortunate :/

PR #653, opened 2025-03-03
feat: better feedback for failed tool calls

### LINKED ISSUE #652: Include all missing fields when tool call fails
Recently I came across this situation: 

```
Calling LLM with the following new messages:
 Assistant: "2. Now, let's update the UserInputCommand parsing:", tools: toolu_01DS8JJg1HMWxhwPLdGx1Mpq#write_file {},
ToolOutput: "toolu_01DS8JJg1HMWxhwPLdGx1Mpq#write_file {}": "Tool call failed: arguments for tool failed to parse: missing field `file_name` at line 1 column 2"
```

Guess what happened? Anthropic then decided to call this tool with the file_name BUT NOT the content. If the error message had stated _all_ the missing fields, I think it may have been able to figure it out. 

In general better responses to failed messages should be given. 

Maybe even easier, the full tool description could be provided once more on each failed tool call

bogzbonny:
closes #652

### DIFF
--- a/swiftide-agents/src/agent.rs
+++ b/swiftide-agents/src/agent.rs
@@ -499,11 +499,21 @@ impl Agent {
                         "Tool call failed, retrying",
                     );
                 }
-                self.add_message(ChatMessage::ToolOutput(
-                    tool_call,
-                    ToolOutput::Fail(error.to_string()),
-                ))
-                .await?;
+                let msg = match self.find_tool_by_name(tool_call.name()) {
+                    Some(tool) => {
+                        // build the message based on the tool description and input_schema
+                        let tool_spec = tool.tool_spec();
+                        let description = tool_spec.description.clone();
+                        let name = tool_spec.name.clone();
+
+                        format!("{error}, please re-read the input_schema from the tools list before reattempting the tool call.\n
+                            tool {name} description: {description}")
+                    }
+                    None => error.to_string(),
+                };
+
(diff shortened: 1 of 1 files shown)


swiftide-agents/src/agent.rs
```diff
@@ -499,11 +499,21 @@ impl Agent {
...
+                    Some(tool) => {
+                        // build the message based on the tool description and input_schema
+                        let tool_spec = tool.tool_spec();
+                        let description = tool_spec.description.clone();
+                        let name = tool_spec.name.clone();
+
+                        format!("{error}, please re-read the input_schema from the tools list before reattempting the tool call.\n
+                            tool {name} description: {description}")
+                    }
+                    None => error.to_string(),
```
timonv, line 512: I'm not sure about this. It could also be other errors. If you want to handle input errors specifically, you can match on the error and tailor the message.

It's also possible to do this in kwaak to experiment a bit with it before committing to a major change with an AfterTool hook, it takes the result &mut

timonv (changes requested):
I'm worried that this might have unexpected side effects with other errors. If you want to format parse errors on the tool calls more specifically, I'd suggest to match on the error and only do it when it's a parse error

bogzbonny:
> I'd suggest to match on the error and only do it when it's a parse error

k interesting - not super invested in the PR so closing for now

PR #664, opened 2025-03-06
fix(integrations): Make openai parallel_tool_calls an Option

bogzbonny:
o3-mini needs to omit parallel_tool_calls - so we need to allow for a None option to not include that field 

needed for https://github.com/bosun-ai/kwaak/pull/392


timonv (approved):
Nice job! Hope this works

PR #684, opened 2025-03-16
fix(integrations): Add chrono/utc feature flag when using qdrant

joshka:
The Qdrant integration calls chrono::Utc::now(), which requires the now feature flag to be enabled in the chrono crate when using qdrant

https://github.com/bosun-ai/swiftide/blob/5f2a5e29b8ba98ddf4fc44f20c1e8da3f25229fa/swiftide-integrations/src/qdrant/indexing_node.rs#L57


timonv (approved):

timonv:
Thanks. There might be a couple more like these after the last release. I'll double check per feature if I missed anything else.

joshka:
> Thanks. There might be a couple more like these after the last release. I'll double check per feature if I missed anything else.

You might consider running [cargo-hack](https://crates.io/crates/cargo-hack) in your CI to catch feature problems.

PR #685, opened 2025-03-16
docs: document feature flags using doc_auto_cfg

joshka:
https://doc.rust-lang.org/rustdoc/unstable-features.html#extensions-to-the-doc-attribute

This is an unstable feature, so it's feature flagged with a feature that
is always passed by docs.rs. To run the same settings use cargo-docs-rs

Also moves the logo into the top left corner of all docs, rather than
just in the main doc.

Makes the docs.rs config consistent across all crate manifests.

### DIFF
--- a/swiftide-agents/Cargo.toml
+++ b/swiftide-agents/Cargo.toml
@@ -34,3 +34,8 @@ insta.workspace = true
 
 [lints]
 workspace = true
+
+[package.metadata.docs.rs]
+all-features = true
+cargo-args = ["-Zunstable-options", "-Zrustdoc-scrape-examples"]
+rustdoc-args = ["--cfg", "docsrs"]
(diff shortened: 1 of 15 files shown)


joshka:
Example:

image

timonv:
> https://doc.rust-lang.org/rustdoc/unstable-features.html#extensions-to-the-doc-attribute
> 
> This is an unstable feature, so it's feature flagged with a feature that is always passed by docs.rs. To run the same settings use cargo-docs-rs
> 
> Also moves the logo into the top left corner of all docs, rather than just in the main doc.
> 
> Makes the docs.rs config consistent across all crate manifests.

Nice! That's very useful. I haven't checked, did it find any examples scraping?

> Also moves the logo into the top left corner of all docs, rather than
just in the main doc.

<3

timonv (approved):

joshka:
> Nice! That's very useful. I haven't checked, did it find any examples scraping?

I think it will only find examples that are under the crate folder, not examples at the root. I mainly copied this section from the main swiftide cargo.toml - good to have if you ever do add examples per crate, but likely doing nothing right now.

PR #692, opened 2025-03-16
docs: wrap code comments at 100 chars

joshka:
This makes it easier to read the source code without horizontal scrolling in github and withing an IDE running on a laptop with a reasonably sized side bar showing

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,5 @@
+# docs: https://rust-lang.github.io/rustfmt/
+
+# Unstable options - to run these, use `cargo +nightly fmt`
+wrap_comments = true
+comment_width = 100
--- a/swiftide-integrations/src/duckdb/mod.rs
+++ b/swiftide-integrations/src/duckdb/mod.rs
@@ -24,7 +24,9 @@ const DEFAULT_UPSERT_QUERY: &str = include_str!("upsert.sql");
 pub struct Duckdb {
     /// The connection to the database
     #[builder(setter(custom))]
-    connection: Arc<Mutex<duckdb::Connection>>, // should be rwlock, however, internally duckdb uses a refcell, so we need the mutex for sync :(
+    connection: Arc<Mutex<duckdb::Connection>>, /* should be rwlock, however, internally duckdb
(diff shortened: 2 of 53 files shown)


rustfmt.toml
joshka, line 1: Only addition in this PR. All other changes in this commit are automatically applied by rustfmt

swiftide-integrations/src/duckdb/mod.rs
```diff
@@ -24,7 +24,9 @@ const DEFAULT_UPSERT_QUERY: &str = include_str!("upsert.sql");
 pub struct Duckdb {
     /// The connection to the database
     #[builder(setter(custom))]
-    connection: Arc<Mutex<duckdb::Connection>>, // should be rwlock, however, internally duckdb uses a refcell, so we need the mutex for sync :(
+    connection: Arc<Mutex<duckdb::Connection>>, /* should be rwlock, however, internally duckdb
+                                                 * uses a refcell, so we need the mutex for sync
+                                                 * :( */
```
joshka: There's a few places like this which are manually cleaned up in the 3rd commit

joshka: Note - RwLock is often not actually better than using a Mutex, and also the std mutex should often be preferred over the tokio one https://tokio.rs/tokio/tutorial/shared-state#on-using-stdsyncmutex-and-tokiosyncmutex (PR to fix this is coming)

timonv: > Note - RwLock is often not actually better than using a Mutex, and also the std mutex should often be preferred over the tokio one https://tokio.rs/tokio/tutorial/shared-state#on-using-stdsyncmutex-and-tokiosyncmutex (PR to fix this is coming)

Shouldn't it be the preferred choice if you want to read in parallel and write locked?

timonv (dismissed):
Neat, that's a really nice improvement

swiftide-integrations/src/duckdb/mod.rs
joshka: Yes. I'm probably misrecalling the exact discussion about this.

PR #693, opened 2025-03-16
chore: use std::sync::Mutex instead of tokio mutex

joshka:
https://tokio.rs/tokio/tutorial/shared-state\#on-using-stdsyncmutex-and-tokiosyncmutex

### DIFF
--- a/swiftide-agents/src/default_context.rs
+++ b/swiftide-agents/src/default_context.rs
@@ -7,14 +7,15 @@
 //! If chat messages include a `ChatMessage::Summary`, all previous messages are ignored except the
 //! system prompt. This is useful for maintaining focus in long conversations or managing token
 //! limits.
-use std::sync::atomic::{AtomicUsize, Ordering};
-use std::sync::Arc;
+use std::sync::{
+    atomic::{AtomicUsize, Ordering},
+    Arc, Mutex,
+};
 
 use anyhow::Result;
 use async_trait::async_trait;
 use swiftide_core::chat_completion::ChatMessage;
 use swiftide_core::{AgentContext, Command, CommandError, CommandOutput, ToolExecutor};
-use tokio::sync::Mutex;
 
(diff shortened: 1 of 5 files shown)


timonv:
I'm aware of the difference, they should be a conscious choice (should != are). Tests that don't require multi-thread / the runtime to yield might give a false positive. I have no idea how well these are covered, perhaps some of the examples can help verify.

joshka:
The tokio version is only required when you need to hold the mutex over an await point. Using std mutexes won't compile in that case. If this compiles then it's correct.

timonv:
> The tokio version is only required when you need to hold the mutex over an await point. Using std mutexes won't compile in that case. If this compiles then it's correct.

Really? I trust your judgement here. I had issues where (I think) the std ones locked and didn't give room to other tasks, switching to a tokio mutex fixed that. No compiler issues and passing tests. I assumed from the docs that a std mutex "locks and blocks", and a tokio mutex "locks, and yields". Is that incorrect? Maybe I was doing something else wrong and a tokio mutex (incorrectly) happened to fix it.

timonv (approved):

timonv (approved):

joshka:
> Really? I trust your judgement here. I had issues where (I think) the std ones locked and didn't give room to other tasks, switching to a tokio mutex fixed that. No compiler issues and passing tests. I assumed from the docs that a std mutex "locks and blocks", and a tokio mutex "locks, and yields". Is that incorrect? Maybe I was doing something else wrong and a tokio mutex (incorrectly) happened to fix it.

From https://tokio.rs/tokio/tutorial/shared-state#on-using-stdsyncmutex-and-tokiosyncmutex

> A synchronous mutex will block the current thread when waiting to acquire the lock. This, in turn, will block other tasks from processing. However, switching to tokio::sync::Mutex usually does not help as the asynchronous mutex uses a synchronous mutex internally.

PR #694, opened 2025-03-16
chore: include filenames in errors in file io

joshka:
Uses fs-err crate to automatically include filenames in the error messages


timonv (approved):
Nice!

Issue #686, opened 2025-03-16
Better logging of pipeline happenings


joshka:
**Is your feature request related to a problem? Please describe.**
I ran my first indexing pipeline using swiftide today and noticed that the level of detail in the log messages was inconsistent in level.

<details><summary>Code</summary>
<p>

```rust
use swiftide::{
    indexing::{
        self,
        loaders::FileLoader,
        transformers::{ChunkCode, Embed, MetadataQACode},
    },
    integrations::{fastembed::FastEmbed, ollama::Ollama, qdrant::Qdrant},
};
use tracing::{info, info_span};
use tracing_subscriber::EnvFilter;

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    dotenvy::dotenv()?;
    let directives = "info,swiftide_indexing=debug,swiftide_integrations=debug";
    let filter = EnvFilter::builder().parse(directives)?;
    tracing_subscriber::fmt().with_env_filter(filter).init();

    let _setup = info_span!("setup").entered();
    info!("Starting swiftide");
    let ollama = Ollama::builder().default_prompt_model("phi4").build()?;
    let loader = FileLoader::new(".").with_extensions(&["rs"]);
    // let redis_url = std::env::var("REDIS_URL")?;
    // let redis = Redis::try_from_url(redis_url, "gusteau")?;
    let rust_chunker = ChunkCode::try_for_language_and_chunk_size("rust", 10..2048)?;
    let fast_embed = FastEmbed::try_default()?;
    let qdrant = Qdrant::builder().batch_size(50).vector_size(1536).build()?;
    info!("Setup finished");
    drop(_setup);

    let _pipeline = info_span!("pipeline").entered();
    info!("Running pipeline");
    indexing::Pipeline::from_loader(loader)
        .with_default_llm_client(ollama)
        // .filter_cached(redis)
        .then_chunk(rust_chunker)
        .then(MetadataQACode::default())
        .then_in_batch(Embed::new(fast_embed))
        .then_store_with(qdrant)
        .run()
        .await?;
    info!("Pipeline finished");

    Ok(())
}
```

```toml
[package]
name = "gusteau"
version = "0.1.0"
edition = "2024"

[dependencies]
anyhow = "1.0.97"
# fix for missing feature flag in swiftide
chrono = { version = "0.4.40", features = ["now"] }
dotenvy = "0.15.7"
swiftide = { version = "0.22.3", features = [
    "fastembed",
    "ollama",
[...]

PR #695, opened 2025-03-17
refactor: improve file loader logging

joshka:
The logging now logs appropriate detail about each file and node processed

- Skipped files are logged at TRACE level
- Configuration of the loader is logged at the start
- The necessary details to identify the loader (path) is included in the span fields
- You can now correlate a node id with a path

image

```
2025-03-17T00:34:18.965601Z DEBUG pipeline:file_loader{root=.}: swiftide_indexing::loaders::file_loader: Loading files extensions=Some(["rs"])
2025-03-17T00:34:18.968115Z TRACE pipeline:file_loader{root=.}:load{path=./Cargo.toml}: swiftide_indexing::loaders::file_loader: Skipping file with extension toml
2025-03-17T00:34:18.968153Z TRACE pipeline:file_loader{root=.}:load{path=./Cargo.lock}: swiftide_indexing::loaders::file_loader: Skipping file with extension lock
2025-03-17T00:34:18.968186Z TRACE pipeline:file_loader{root=.}:load{path=./README.md}: swiftide_indexing::loaders::file_loader: Skipping file with extension md
2025-03-17T00:34:18.968248Z DEBUG pipeline:file_loader{root=.}:load{path=./src/main.rs}: swiftide_indexing::loaders::file_loader: Loading file
2025-03-17T00:34:18.968302Z DEBUG pipeline:file_loader{root=.}:load{path=./src/main.rs}: swiftide_indexing::loaders::file_loader: Loaded file node_id=9cef7b97-dfbb-327b-a47d-0816f4ed69db
2025-03-17T00:34:18.968324Z TRACE pipeline:file_loader{root=.}:load{path=./src/compose.yaml}: swiftide_indexing::loaders::file_loader: Skipping file with extension yaml
```

Related to https://github.com/bosun-ai/swiftide/issues/686

### DIFF
--- a/swiftide-indexing/src/loaders/file_loader.rs
+++ b/swiftide-indexing/src/loaders/file_loader.rs
@@ -5,24 +5,25 @@ use std::{
 };
 
 use anyhow::Context as _;
+use ignore::{DirEntry, Walk};
 use swiftide_core::{indexing::IndexingStream, indexing::Node, Loader};
+use tracing::{debug_span, instrument, Span};
 
-/// The `FileLoader` struct is responsible for loading files from a specified directory,
-/// filtering them based on their extensions, and creating a stream of these files for further
-/// processing.
+/// The `FileLoader` struct is responsible for loading files from a specified directory, filtering
+/// them based on their extensions, and creating a stream of these files for further processing.
 ///
 /// # Example
 ///
+/// Create a pipeline that loads the current directory and indexes all files with the ".rs"
+///
 /// ```no_run
-/// // Create a pipeline that loads the current directory
-/// // and indexes all files with the ".rs" extension.
 /// # use swiftide_indexing as indexing;
 /// # use swiftide_indexing::loaders::FileLoader;
 /// indexing::Pipeline::from_loader(FileLoader::new(".").with_extensions(&["rs"]));
 /// ```
 #[derive(Clone, Debug)]
 pub struct FileLoader {
-    pub(crate) path: PathBuf,
+    pub(crate) root: PathBuf,
     pub(crate) extensions: Option<Vec<String>>,
 }
 
@@ -30,13 +31,15 @@ impl FileLoader {
     /// Creates a new `FileLoader` with the specified path.
     ///
     /// # Arguments
-    /// * `path` - The path to the directory to load files from.
+    ///
+    /// - `root`: The root directory to load files from.
     ///
     /// # Returns
+    ///
     /// A new instance of `FileLoader`.
-    pub fn new(path: impl Into<PathBuf>) -> Self {
+    pub fn new(root: impl AsRef<Path>) -> Self {
         Self {
-            path: path.into(),
+            root: root.as_ref().to_path_buf(),
             extensions: None,
         }
     }
@@ -44,28 +47,27 @@ impl FileLoader {
     /// Adds extensions to the loader.
     ///
     /// # Arguments
-    /// * `extensions` - A list of extensions to add without the leading dot.
+    ///
+    /// - `extensions`: A list of extensions to add without the leading dot.
     ///
     /// # Returns
+    ///
     /// The `FileLoader` instance with the added extensions.
     #[must_use]
     pub fn with_extensions(mut self, extensions: &[impl AsRef<str>]) -> Self {
-        self.extensions = Some(
-            self.extensions
-                .unwrap_or_default()
-                .into_iter()
-                .chain(extensions.iter().map(|ext| ext.as_ref().to_string()))
-                .collect(),
-        );
+        let existing = self.extensions.get_or_insert_default();
+        existing.extend(extensions.iter().map(|ext| ext.as_ref().to_string()));
         self
     }
 
     /// Lists the nodes (files) that match the specified extensions.
     ///
     /// # Returns
+    ///
     /// A vector of `Node` representing the matching files.
     ///
     /// # Panics
+    ///
     /// This method will panic if it fails to read a file's content.
     pub fn list_nodes(&self) -> Vec<Node> {
         self.iter().filter_map(Result::ok).collect()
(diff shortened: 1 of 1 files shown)


swiftide-indexing/src/loaders/file_loader.rs
```diff
@@ -5,118 +5,181 @@ use std::{
...
     /// # Arguments
-    /// * `path` - The path to the directory to load files from.
+    ///
+    /// - `root`: The root directory to load files from.
     ///
     /// # Returns
+    ///
     /// A new instance of `FileLoader`.
-    pub fn new(path: impl Into<PathBuf>) -> Self {
+    pub fn new(root: impl AsRef<Path>) -> Self {
```
joshka, line 40: This aligns more with the params you see across the std lib

swiftide-indexing/src/loaders/file_loader.rs
```diff
@@ -5,118 +5,181 @@ use std::{
...
-
-                let original_size = content.len();
-
-                Node::builder()
-                    .path(entry.path())
-                    .chunk(content)
-                    .original_size(original_size)
-                    .build()
-            })
+        Iter::new(&self.root, self.extensions.clone()).fuse()
```
joshka, line 78: The fuse here means that the iterator won't be called again once complete. This might be useful if we want to make the last item spit out stats or something similar, though those could also just be in a drop impl maybe?

swiftide-indexing/src/loaders/file_loader.rs
```diff
@@ -5,118 +5,181 @@ use std::{
...
+            }
+        }
+    }
+}
+
+impl Iter {
+    /// Creates a new `Iter` instance.
+    fn new(root: &Path, include_extensions: Option<Vec<String>>) -> Self {
+        let span = debug_span!("file_loader", root = %root.display());
+        tracing::debug!(parent: &span, extensions = ?include_extensions, "Loading files");
```
joshka, line 122: I didn't put the extensions in the span fields because it probably only really needs to be logged once

swiftide-indexing/src/loaders/file_loader.rs
```diff
@@ -5,118 +5,181 @@ use std::{
...
+            walk: Walk::new(root),
+            include_extensions,
+            span,
+        }
+    }
+
+    #[instrument(skip_all, fields(path = %entry.path().display()))]
+    fn load(&self, entry: DirEntry) -> Option<anyhow::Result<Node>> {
+        if entry.file_type().is_some_and(|ft| !ft.is_file()) {
+            // Skip directories and non-files
```
joshka, line 133: not logging for non files - just skipping them. This could do a "recursing into director foo" type thing at some point in the future if needed.

swiftide-indexing/src/loaders/file_loader.rs
```diff
@@ -5,118 +5,181 @@ use std::{
...
+            return None;
+        }
+        if let Some(extensions) = &self.include_extensions {
+            let Some(extension) = entry.path().extension() else {
+                tracing::trace!("Skipping file without extension");
```
joshka, line 138: These are verbose, but the sort of verbose that you enable because there's some stupid problem you're bashing your head against in your code which is immediately obvious when turned on.

swiftide-indexing/src/loaders/file_loader.rs
```diff
@@ -5,118 +5,181 @@ use std::{
...
+        }
+        tracing::debug!("Loading file");
+        match read_node(entry) {
+            Ok(node) => {
+                tracing::debug!(node_id = %node.id(), "Loaded file");
+                Some(Ok(node))
+            }
+            Err(err) => {
+                tracing::error!(error = %err, "Failed to load file");
+                Some(Err(err))
```
joshka, line 155: Generally I avoid doing both log + return error, so perhaps this should be skipped. However looking at the current handling of the error, it doesn't seem like it currently gets an error log message, instead it's passed back to the caller. WDYT?

joshka (commented):
Added some comments to help focus the review where I think there's most interesting decisions.

swiftide-indexing/src/loaders/file_loader.rs
timonv: Yeah, currently it's up to the pipeline implementer to handle, log or ignore it. So each failed read bubbles up in the pipeline. If you do not handle it, calling query on the pipeline will result in an error. There's a couple of helpers for that in the pipeline itself. Kwaak blatantly logs and ignores them.

With that, should we remove the log here? Or do you think we should accept a bit of duplication here to ease users into it?

swiftide-indexing/src/loaders/file_loader.rs
timonv: With that, wouldn't you want it to return None once and then stop yielding all together?

I'm tempted to have stats (or something similar) be something generically publish to something else that you hook into a pipeline. Not all loaders finish, or might take a very long time to run.

timonv (commented):
Nice, this looks really good. I just want to clarify up one thing on the fuse call and align on error logging.

swiftide-indexing/src/loaders/file_loader.rs
joshka: This is what it would look like if I added the following to my calling code to log pipeline errors:
```rust
    {
        Ok(_) => info!("Pipeline finished successfully"),
        Err(e) => {
            tracing::error!(error = ?e, "Pipeline failed");
            return Err(e);
        }
    }
```

image

The good thing about having the error logged here is that it carries the context of which loader / file has the problem.

I think perhaps leave the log message here, but keep it in mind for removal when thinking holistically about how error propagation works (i.e. when you eventually ditch anyhow and error types).

swiftide-indexing/src/loaders/file_loader.rs
joshka: I think you definitely want an iterator to continue to respond with None when done.

This is effectively a noop right now as the code which returns None does nothing. What it does do is protect the `Walk` iterator from being called again once it's returned None. I didn't see any specific wording on the ignore crate docs about what calling `.next()` after completion will return (this is generally undefined behavior based on the iterator trait). I'm assuming that it's likely that it works fine and does nothing, but this makes it clear that we don't need to rely on assumption and we don't have to test it on various different operating systems etc.

The upshot is this doesn't break anything, so I'd suggest keep it. I'd maybe add a small comment, but I can't think of one which does more than repeat what fuse does (prevents calling the iterator again once it's finished).

> I'm tempted to have stats (or something similar) be something generically publish to something else that you hook into a pipeline (ie with a channel). Not all loaders finish, or might take a very long time to run.

Yeah. I think that that would be a good idea. Incremental stats makes for good interactive UIs.

joshka:
Rebased on main

swiftide-indexing/src/loaders/file_loader.rs
timonv: Ok perfect, thanks for clearing that up.

timonv (approved):

PR #739, opened 2025-04-14
feat: Multi agent setups with tasks

timonv:
- **wip**
- **wip**
- **wip**
- **Implement adding delegate tool**
- **more wip**


swiftide-agents/src/agent.rs
```diff
@@ -116,6 +119,33 @@ pub struct Agent {
...
+
+/// Cloning an agent is expensive
+///
+/// It's intended use is to have an identically configured agent
+/// operating on a new context. This allows spawning many agents from a single configuration.
+impl Clone for Agent {
+    fn clone(&self) -> Self {
+        Agent {
+            hooks: self.hooks.clone(),
+            context: Arc::new(self.context.clone_empty()),
```
timonv, line 136: Maybe not change this behaviour and add a clone_empty instead to the agent

swiftide-agents/src/tasks/task.rs
```diff
@@ -0,0 +1,233 @@
...
+//
+#[derive(Builder, Clone, Debug)]
+#[builder(build_fn(skip, error = TaskBuilderError))]
+pub struct Task {
+    #[builder(field(ty = "Option<Vec<RunningAgent>>"))]
+    agents: Arc<tokio::sync::RwLock<Vec<RunningAgent>>>,
+    #[builder(field(ty = "Option<Vec<Action>>"))]
+    actions: Arc<Vec<Action>>,
+    starts_with: Arc<String>,
+    state: Arc<TaskState>,
```
timonv, line 51: Perhaps make the task generic over a state instead that can be used in hooks etc. Not sure if this is needed.

swiftide-agents/src/tasks/running_agent.rs
```diff
@@ -0,0 +1,40 @@
...
+//!
+//! This allows
+
+use std::{ops::Deref, sync::Arc};
+
+use crate::Agent;
+#[derive(Clone, Debug)]
+pub(crate) struct RunningAgent(Arc<tokio::sync::Mutex<Agent>>, Arc<String>);
+
+impl RunningAgent {
```
timonv, line 14: Maybe name it a `TaskAgent` instead. Technically it's not 'running', it's just wrapped with inner mutability and a ref for easy access.

swiftide-agents/src/tasks/task.rs
```diff
@@ -0,0 +1,233 @@
...
+    #[builder(private, default)]
+    current_agent: Arc<AtomicUsize>,
```
timonv, line 53: Conceptually this (might) breaks when running multiple agents in parallel

swiftide-agents/src/tasks/task.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    #[builder(private, default = Arc::new(Mutex::new(JoinSet::new())))]
+    // All spawned agents
+    running_agents: Arc<std::sync::Mutex<JoinSet<Result<(), TaskError>>>>,
```
timonv, line 57: Perhaps add a limit to concurrency right away. Nobody likes a fork bomb.

nikhil-pandey:
A few thoughts:
- **Composable workflows**: Could a `Task` itself be treated as an action that an agent can run? That’d enable recursive composition (e.g., agents executing task graphs that themselves contain subtasks).
- **Workflow coverage**: Curious if you’re aiming to support the common agentic workflows Anthropic outlines [here](https://www.anthropic.com/engineering/building-effective-agents) — would be great if we can add examples.
- **Durability**: Have you thought about resumability/persistence for long-running or restartable tasks? Especially useful for more autonomous or stateful agent flows.

Also, do you have a doc or diagram sketching out where this is going long-term? Would love to get a better sense of how you see the agent/task system evolving.

timonv:
Hey @nikhil-pandey,

Thanks, this is really good feedback, I appreciate it!

> A few thoughts:
> 
>     * **Composable workflows**: Could a `Task` itself be treated as an action that an agent can run? That’d enable recursive composition (e.g., agents executing task graphs that themselves contain subtasks).

Sure, why not, you can go as wild as you want. I'm not building it with that intention, but it should certainly be possible.

>     * **Workflow coverage**: Curious if you’re aiming to support the common agentic workflows Anthropic outlines [here](https://www.anthropic.com/engineering/building-effective-agents) — would be great if we can add examples.

My vision is a bit of a hybrid. I like the idea of agents acting autonomously, and then using hooks at certain lifecycle events to control the flow. Swiftide agents already have a lot of that, and then the idea is that tasks facilitate more complicated multi agent flows.

>     * **Durability**: Have you thought about resumability/persistence for long-running or restartable tasks? Especially useful for more autonomous or stateful agent flows.

Yeah. That would be really nice. The default context and message history is fully serializable, so it's actually not that far off. For tasks it's a bit more involved, but I recon that same direction can be used.

>     * **Dynamic dispatch**: Would love to see support for agents or even tasks being _exposed as tools_ to other agents — making it possible to treat an entire workflow as a callable operation in a higher-level orchestration.

You can create a tool that takes an agent or a task. I'm doing something similar with the delegate tool. In a later release, I'd also want to provide an api so it's easy for agents to spawn other agents (accounting for fork bombs).

> Also, do you have a doc or diagram sketching out where this is going long-term? Would love to get a better sense of how you see the agent/task system evolving.

Not quite yet. The idea for this initial release is to facilitate delegation (so agents run in sequence), with hooks on important lifecycle events. Then in a later release add the option for agents to operate in parallel. I think the architecture already facilitates that.

I really appreciate the feedback, keep it coming 🙏

nikhil-pandey:
👋 Just a quick heads-up — Temporal recently [landed durability support](https://github.com/temporalio/samples-python/commit/14d1e61648388c0e2b08b55c4b9689f4c599df36) for OpenAI's agents SDK in their Python samples. I gave it a spin, and it works really well. After each model completion or tool call, Temporal persistently stores the interaction state — making long-running, resumable agent workflows seamless out of the box.

It would be awesome if we could support a similar experience via the [Temporal Rust SDK](https://github.com/temporalio/sdk-core/blob/master/sdk/src/lib.rs). Given how nicely Swiftide tasks and agent state are structured, this feels like a natural next step — especially for enabling reliable, stateful orchestration of multi-agent systems.

Just worth noting: the Rust SDK isn’t officially public-facing (yet) — it’s what powers Temporal’s polyglot SDKs internally — but their core execution engine is rock solid and a perfect fit for this kind of use case.

PR #740, opened 2025-04-15
fix(agents): Properly support nullable types for MCP tools

timonv:


timonv:
Hmm after some more testing, the call works, but the json spec is incorrect. Seems the serializer does not get called at all for `Nullable`, but for `String` instead. Maybe it's related to https://github.com/serde-rs/serde/issues/1183, will need some time to investigate.

timonv:
Should be fixed, the generated spec looks correct.

PR #745, opened 2025-04-15
chore: Pretty print RAGAS output

tinco:
It's a bit unreadable without the formatting.


timonv:
Doesn't jq work? If you want pretty serde_json does have a helper

timonv:
Fine by me as long as no one complains 😅

PR #790, opened 2025-05-13
feat(agents)!: Easy human-in-the-loop flows by decorating tools

timonv:
Tools can now return a FeedbackRequired with an optional Value payload. The agent will stop, with the tool call and optional payload in the reason. Possible to use either the stop reason returned by the agent or a hook to act on this, or use it in your own tools.

Feedback is given on the context of the agent with the tool call and an optional payload. The payload offers flexibility to provide feedback to the tool, or to provide context to the user.

BREAKING CHANGE: The `Tool` trait now receives a `ToolCall` as argument instead of an `Option<&str>`. The latter is still accessible via `tool_call.args()`.

### DIFF
--- a/examples/agents_with_human_in_the_loop.rs
+++ b/examples/agents_with_human_in_the_loop.rs
@@ -21,14 +21,6 @@ use swiftide::{
     traits::{AgentContext, ToolFeedback},
 };
 
-// The macro supports strings/strs, vectors/slices, booleans and numbers.
-//
-// This is currently only supported for the attribute macro, not the derive macro.
-//
-// If you need more control or need to use full objects, we recommend to implement the `Tool` trait
-// and prove the Json spec yourself. Builders are available.
-//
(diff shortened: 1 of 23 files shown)


examples/agents_with_human_in_the_loop.rs
```diff
@@ -0,0 +1,105 @@
...
+    chat_completion::{ToolCall, ToolOutput, errors::ToolError},
+    traits::AgentContext,
+};
+
+// The macro supports strings/strs, vectors/slices, booleans and numbers.
+//
+// This is currently only supported for the attribute macro, not the derive macro.
+//
+// If you need more control or need to use full objects, we recommend to implement the `Tool` trait
+// and prove the Json spec yourself. Builders are available.
```
timonv, line 28: docs: Add that if you implement your own workflow, there is no need to deal with the hook. Additionally, the tool call is also in the stop reason, the hook is just another example.

PR #800, opened 2025-05-25
fix(pipeline): Cache nodes after they've been successfully ran

adrianncovaci:
See #151

### DIFF
--- a/swiftide-core/src/node.rs
+++ b/swiftide-core/src/node.rs
@@ -63,6 +63,9 @@ pub struct Node {
     /// from in bytes
     #[builder(default)]
     pub offset: usize,
+    /// Optional parent id
+    #[builder(default)]
+    pub parent_id: Option<uuid::Uuid>,
 }
 
 impl NodeBuilder {
@@ -126,7 +129,7 @@ impl Debug for Node {
 impl Node {
     /// Builds a new instance of `Node`, returning a `NodeBuilder`. Copies
     /// over the fields from the provided `Node`.
-    pub fn build_from_other(node: &Node) -> NodeBuilder {
+    pub fn chunking_from(node: &Node) -> NodeBuilder {
         NodeBuilder::default()
             .path(node.path.clone())
             .chunk(node.chunk.clone())
@@ -136,6 +139,7 @@ impl Node {
             .embed_mode(node.embed_mode)
             .original_size(node.original_size)
             .offset(node.offset)
+            .parent_id(node.id())
             .to_owned()
     }
 
@@ -241,6 +245,10 @@ impl Node {
 
         uuid::Uuid::new_v3(&uuid::Uuid::NAMESPACE_OID, &bytes)
     }
+
+    pub fn parent_id(&self) -> Option<uuid::Uuid> {
+        self.parent_id
+    }
 }
 
 impl Hash for Node {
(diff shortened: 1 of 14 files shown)


timonv (changes requested):
Thank you for the contribution!

Overal, I have three concerns:
- This will not work when nodes are chunked/split, as the parent node won't persist through the stream, only its children.
- I'm worried the mutex approach will bottleneck the stream
- The clone on the node keeps the node for a longer duration in memory, this might be negligible, I'm not sure.

One alternative approach could be to expand the nodecache trait so that the stream can get the key, then use a channel to store the node id / key. On stream completion it could spawn a task to persist the cache. When splitting, nodes would need to keep track of their parent id (the node_id). Which is useful either way.

That's just one solution, what are your thoughts?

swiftide-core/src/node.rs
```diff
@@ -63,6 +63,9 @@ pub struct Node {
     /// from in bytes
     #[builder(default)]
     pub offset: usize,
+    /// Optional parent id
+    #[builder(default)]
+    pub origin_id: Option<uuid::Uuid>,
```
timonv, line 68: Could you rename this to `parent_id`? I think that fits the comment and usecase better

timonv: Additionaly, instead of using the builder directly, how about renaming `build_from_other` to `chunking_from`, then also setting the parent_id. Afaik it's not used anywhere else, makes it a lot more clear, and removes the need for calling the builder method.

timonv (changes requested):
Nice! This solution looks really good.

Current caches (Redis, Redb, Duckdb), use the `node.id()` to generate the cache key. I think the block at 184 will fail.

What I'd suggest is to add an integration test in swiftide/tests, with a controlled amount of nodes (you can create a new pipeline from an `IndexingStream`, and the latter can be created from anything that iterates over `Result<Node>`.

Suspect something like:
```rust
let nodes = vec![node1,node2,node3];
Pipeline::from_stream(nodes);
```
will work. There's also some other helpers, been a while since I touched that and looking at that now it coulde be improved 😅 

Then in the pipeline, touch everything changed here, caching, chunking, splitting, merging and storing.

Finally, assert that the initial nodes were cached. That should quickly show if there are any problems. No mocks needed.

Thanks for the hard work! Let me know if there is anything I can help with, feel free to DM me on discord. 🙏

swiftide-indexing/src/pipeline.rs
```diff
@@ -170,6 +180,32 @@ impl Pipeline {
...
+                cache_for_task.set(&cache_node).await;
+            }
+        });
+
+        self.cache_handle = Some(
+            handle
+                .map(|result| {
+                    if let Err(err) = result {
+                        // TOCHECK: Should we continue or panic here?
+                        tracing::error!("Cache task failed: {err}");
```
timonv, line 202: I think logging is enough? I'd rather avoid panics if possible. Alternative would be to propagate the error back to the main stream

swiftide-indexing/src/pipeline.rs
```diff
@@ -595,19 +635,30 @@ impl Pipeline {
             .collect::<Vec<_>>();
         futures_util::future::try_join_all(setup_futures).await?;
 
-        let mut total_nodes = 0;
-        while self.stream.try_next().await?.is_some() {
-            total_nodes += 1;
+        let mut cache_keys = HashSet::new();
+
+        while let Some(node) = self.stream.try_next().await? {
+            cache_keys.insert(node.parent_id().unwrap_or(node.id()));
```
timonv, line 641: I'd suggest to send it to the cache if the key has not been seen yet, then they can be processed right away instead of waiting for the stream to complete (you could also handle it there). Counting the cache keys as the node count is not going to work, if there is no cache.

timonv (changes requested):
Excellent, functionally this looks great!

Just one point on counting the total processed nodes, and handing it to the cache task directly instead of waiting for the main pipeline to finish.

swiftide-indexing/src/pipeline.rs
adrianncovaci: Good spot, thank you!

Please let me know if you have any other suggestions.

timonv:
@adrianncovaci Could you rerequest the review when you're ready? Not sure if you're still fixing things 🙏

timonv (dismissed):

PR #830, opened 2025-06-25
feat(qdrant): Add support for a filter in hybrid search

timonv:


Lawouach:
Hey @timonv 

It seems that to acces use qdrant_client::qdrant::Filter; I had to add `qdrant_client` directly. Maybe switfide could re-export qdrant::Filter?

PR #1069, opened 2026-04-22
fix: send error ToolOutput when tool is not found instead of silently skipping

mkg20001:
## Summary

When the LLM calls a tool that isn't registered in the agent's toolbox, `invoke_tools` currently logs a warning and `continue`s without adding a `ToolOutput` to the conversation history. This breaks the `tool_use`/`tool_result` contract that LLM APIs require (Anthropic in particular), causing the next completion request to fail or the agent to terminate unexpectedly.

This PR:
- Sends a `ToolOutput::fail()` back to the conversation when a tool is not found, so the LLM gets explicit feedback and can adapt (e.g. use a different tool)
- Adds a test verifying the agent recovers from an unknown tool call

## Before

```
WARN agent: Tool service-restart not found
// No ToolOutput added → next API call fails or agent stops
```

## After

```
WARN agent: Tool service-restart not found
// ToolOutput::fail("Tool 'service-restart' is not available...") added
// → LLM sees the error and can try a different approach
```

## Test plan

- [x] New test `test_tool_not_found_sends_error_response` — verifies agent sends error and continues
- [x] All existing agent tests pass (63/63 + 1 new, the `test_local_executor_clear_env` failure is pre-existing and unrelated)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


timonv (changes requested):
@mkg20001 Thanks for the contribution!

The reason this was added is the same, if no tool call is present in the history, the contract will also break. IIRC* if you remove the tool, but not the tool call from history, the contract is also invalid. Not sure if openai/anthropic have parity here.

Maybe the fix is a bit more involved, checking for partial and all cases. What do you think?

_* wouldn't put it past me to misremember_

mkg20001:
I'm not sure, I think this depends heavily on what the LLM or backend does with it.

This is what claude had to say about it: https://claude.ai/share/419bfb11-b9fa-4417-9a95-ce09ef0e289e

Seems model dependent. I don't know how to proceed, removing it may remove important context and make the model behave differently, but it also risks causing further problems with _some_ models. Can we get this PR in and then patch this in another PR once we figure out how to fix it? Or would it make sense to use a flag with a sensible default (strip by default, toggleable to keep)

timonv:
> I'm not sure, I think this depends heavily on what the LLM or backend does with it.
> 
> This is what claude had to say about it: https://claude.ai/share/419bfb11-b9fa-4417-9a95-ce09ef0e289e
> 
> Seems model dependent. I don't know how to proceed, removing it may remove important context and make the model behave differently, but it also risks causing further problems with _some_ models. Can we get this PR in and then patch this in another PR once we figure out how to fix it? Or would it make sense to use a flag with a sensible default (strip by default, toggleable to keep)

Well, we're using this in production, so I want to be sure it works :') A flag with the current behavior as default sounds like a good idea!

timonv:
@mkg20001 I must have been extremely sleep deprived 😅 This is actually perfect. Sorry for the late response. It's an edge case and you're right, harnass should always attempt to keep tools / calls aligned.

timonv (approved):

PR #1079, opened 2026-05-01
fix(integrations): stop selecting tls providers

tinco:
## Reason

Quak installs the workspace rustls crypto provider at runtime, but `swiftide-integrations` was still selecting provider-bearing TLS features in its integration graph. In the OpenAI/tiktoken path this also kept an older `async-openai` / `reqwest` 0.11 / `rustls` 0.21 lane alive, which adds extra TLS and crypto crates to cold builds.

## What changed

- Switch the workspace `reqwest` feature from `rustls` to `rustls-no-provider` so Swiftide does not choose a crypto provider for downstream applications.
- Add a private `reqwest` 0.12 dependency using `rustls-tls-native-roots-no-provider` for the OpenAI and Anthropic feature paths.
- Stop enabling `tiktoken-rs?/async-openai`, which removes the old `async-openai` 0.14 / `reqwest` 0.11 / `rustls` 0.21 lockfile lane.
- Update `Cargo.lock` after the dependency graph cleanup and the rebase onto current `master`.

## Impact

Swiftide still uses rustls for these integration paths, but provider selection is now left to the application or workspace. Consumers that enable these features must continue to install their chosen rustls provider at runtime, such as `aws_lc_rs::default_provider().install_default()` in quak.

## Benefits

- Keeps AWS-LC as an application-level provider choice instead of selecting another provider inside Swiftide.
- Removes the older reqwest/rustls lane from the OpenAI/tiktoken feature graph.
- Reduces duplicated TLS and crypto dependencies, which should improve clean build times for quak and other downstream workspaces that already normalize on a rustls provider.

## Package version check

- `async-anthropic` is already on the latest crates.io release, `0.6.0`.
- `reqwest-eventsource` is already on the latest crates.io release, `0.6.0`.
- The workspace `reqwest` 0.13 line now resolves to `0.13.3` in the rebased lockfile.
- `async-openai` latest is `0.36.1`, but it changes response/tool types that Swiftide re-exports or adapts. This PR keeps `async-openai` on the compatible `0.33.0` lockfile version to avoid interface or adapter changes.
- `tiktoken-rs` latest is `0.11.0`, but it also requires adapter changes. This PR keeps the compatible `0.9.1` lockfile version.

## Validation

- `git diff --check`
- `cargo check --locked -p swiftide-integrations --features openai`
- `cargo check --locked -p swiftide-integrations --no-default-features --features openai,tiktoken,anthropic`
- `cargo tree --locked -p swiftide-integrations --edges normal --features openai -i ring` prints nothing
- `cargo tree --locked -p swiftide-integrations --edges normal --features openai,tiktoken -i reqwest@0.11.27` no longer matches a package
- `cargo tree --locked -p swiftide-integrations --all-features --edges normal -i rustls@0.21.12` no longer matches a package
- `cargo tree --locked --workspace --all-features --edges normal -i openssl-sys` prints nothing on the host target
- `cargo tree --locked -p swiftide-integrations --target x86_64-unknown-linux-gnu --features openai --edges normal -i openssl-sys` prints nothing

## Notes

Linux all-feature builds still have optional native TLS/OpenSSL holdouts through other integration features. This PR keeps the scope to the provider selection and old reqwest lane that affect the quak/OpenAI path.

### DIFF
--- a/swiftide-integrations/Cargo.toml
+++ b/swiftide-integrations/Cargo.toml
@@ -39,7 +39,6 @@ schemars.workspace = true
 
 # Integrations
 async-openai = { workspace = true, optional = true, features = [
-  "rustls",
   "chat-completion",
   "embedding",
   "responses",
@@ -101,6 +100,9 @@ aws-smithy-types = { workspace = true, optional = true }
 aws-smithy-json = { version = "0.62.4", optional = true }
 secrecy = { workspace = true, optional = true }
 reqwest = { workspace = true, optional = true }
+reqwest_0_12 = { package = "reqwest", version = "0.12", default-features = false, features = [
+  "rustls-tls-native-roots-no-provider",
+], optional = true }
 reqwest-eventsource = { workspace = true, optional = true }
 deadpool = { workspace = true, features = [
   "managed",
@@ -158,8 +160,9 @@ tracing-subscriber = { workspace = true }
 default = ["rustls"]
 
 metrics = ["dep:metrics", "swiftide-core/metrics"]
-# Ensures rustls is used
-rustls = ["reqwest?/rustls", "fastembed?/hf-hub-native-tls"]
+# Ensures rustls is used without selecting a crypto provider. Applications can install their
+# workspace provider at runtime, e.g. aws_lc_rs::default_provider().install_default().
+rustls = ["reqwest?/rustls-no-provider", "fastembed?/hf-hub-native-tls"]
 # Qdrant for storage
 qdrant = ["dep:qdrant-client", "swiftide-core/qdrant", "chrono/now"]
 # PgVector for storage
@@ -187,7 +190,7 @@ tree-sitter = [
 # OpenAI for embedding and prompting
 openai = [
   "dep:async-openai",
-  "tiktoken-rs?/async-openai",
+  "dep:reqwest_0_12",
   "dep:reqwest-eventsource",
   "dep:reqwest",
   "swiftide-core/openai",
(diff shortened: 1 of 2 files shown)


tinco:
@timonv I'm not very familiar with how this works, does this make sense at all? I noticed that Quak is spending 1/3rd of the compile time on compiling OpenSSL/RustTLS, and Codex suggested we could make swiftide more flexible allowing us to eliminate some of the crypto compile time.

swiftide-integrations/Cargo.toml
```diff
@@ -101,6 +100,9 @@ aws-smithy-types = { workspace = true, optional = true }
 aws-smithy-json = { version = "0.62.4", optional = true }
 secrecy = { workspace = true, optional = true }
 reqwest = { workspace = true, optional = true }
+reqwest_0_12 = { package = "reqwest", version = "0.12", default-features = false, features = [
```
tinco, line 103: what's the idea behind introducing what I suppose is a virtual package that's an alias for reqwest 0.12 and the specific feature enabled?

timonv: Honestly no clue, some legacy? Don't know when it got introduced, imo shouldn't be a thing

timonv:
Nice, good stuff

timonv (dismissed):

swiftide-integrations/Cargo.toml
timonv: Ah because of anthropic. Honestly, we haven't updated that in ages :')

PR #1125, opened 2026-07-27
feat(scraping): optional Spider Cloud support

j-mendez:
Adds optional [Spider Cloud](https://spider.cloud) support to `ScrapingLoader` for sites that block direct fetches.

```bash
SPIDER_CLOUD_API_KEY=sk-...   # opt-in
SPIDER_CLOUD_MODE=smart       # optional: smart (default) | proxy | unblocker | api | fallback
```

With no key set the loader behaves exactly as before. Applied in `into_stream` so it covers `from_url`, `from_spider`, and the builder.

Loopback and private hosts skip the cloud since it can't reach them — without that the existing wiremock tests fail when a key happens to be set in your environment.

Verified against a live site: cloud path returns content, an invalid key returns nothing (so the proxy is really in the path), and the wiremock tests pass with and without a key set.

Stacks on #1124.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -846,15 +846,14 @@ checksum = "1505bd5d3d116872e7271a6d4e16d81d0c8570876c8de68093a09ac269d8aac0"
 
 [[package]]
 name = "auto_encoder"
-version = "0.1.10"
+version = "0.2.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "3f6364e11e0270035ec392151a54f1476e6b3612ef9f4fe09d35e72a8cebcb65"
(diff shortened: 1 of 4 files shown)


timonv (approved):
Thanks for the contrib, smart move 😎 I suppose that technically this might have issues with private dns? But if that's the case, would be weird to use an external crawler anyway.

j-mendez:
> Thanks for the contrib, smart move 😎 I suppose that technically this might have issues with private dns? But if that's the case, would be weird to use an external crawler anyway.

Np, yes unless we use the remote proxy option for the one off cases.

timonv:
@j-mendez CI fails because of ethnum, which was introduced in this branch. Could you take a look?

timonv:
Jokes on me, must be a recent failure, I'll clean it up. Thanks for the PR 🙏

PR #1146, opened 2026-08-03
feat(core)!: separate command output streams

timonv:
## What changed

- store command output as ordered, tagged `Bytes` chunks
- expose stdout and stderr as iterators over borrowed `Bytes`
- keep combined bytes and text available for existing callers
- remove stream view types and per-stream string helpers
- let the local executor stream pipe reads directly into output chunks

## Why

One ordered chunk list retains the output an agent sees while preserving which pipe produced each chunk. Stream access borrows existing bytes. Combined access borrows empty and single-chunk output and joins multi-chunk output once.

Ordering reflects when the executor observes each pipe. Separate operating-system pipes cannot prove process write order.

## Developer impact

Use `chunks()` for tagged output, `stdout()` or `stderr()` for one pipe, and `as_bytes()`, `to_string_lossy()`, or their consuming forms for combined output.

```rust
for bytes in output.stderr() {
    handle(bytes);
}
```

## Validation

- `cargo +nightly fmt --all -- --check`
- `cargo clippy -p swiftide-core -p swiftide-agents --all-targets --all-features --locked -- -D warnings`
- `cargo test -p swiftide-core command_output --lib --locked`
- `cargo test -p swiftide-agents local_executor --lib --locked`
- `cargo test -p swiftide-core --doc`

### DIFF
--- a/swiftide-core/src/chat_completion/tools.rs
+++ b/swiftide-core/src/chat_completion/tools.rs
@@ -110,7 +110,7 @@ impl<S: AsRef<str>> From<S> for ToolOutput {
 
 impl From<CommandOutput> for ToolOutput {
     fn from(output: CommandOutput) -> Self {
-        ToolOutput::Text(output.into_string_lossy())
+        ToolOutput::Text(output.into_stdout_string_lossy())
     }
 }
 impl std::fmt::Display for ToolOutput {
--- a/swiftide-core/src/chat_completion/tools.rs
+++ b/swiftide-core/src/chat_completion/tools.rs
@@ -110,9 +110,35 @@ impl<S: AsRef<str>> From<S> for ToolOutput {
 
 impl From<CommandOutput> for ToolOutput {
     fn from(output: CommandOutput) -> Self {
-        ToolOutput::Text(output.into_stdout_string_lossy())
+        ToolOutput::Text(output.into_string_lossy())
     }
 }
+
+#[cfg(test)]
+mod command_output_tests {
+    use super::{CommandOutput, ToolOutput};
+    use crate::CommandOutputChunk;
+    use bytes::Bytes;
+
+    #[test]
+    fn keeps_stdout_and_stderr_in_tool_output() {
+        let output = CommandOutput::from_chunks([
+            CommandOutputChunk::Stdout(Bytes::from_static(b"out")),
+            CommandOutputChunk::Stderr(Bytes::from_static(b"err")),
+        ]);
+
+        assert_eq!(ToolOutput::from(output), ToolOutput::Text("outerr".into()));
+    }
+
+    #[test]
+    fn keeps_stderr_only_tool_output() {
+        let output =
+            CommandOutput::from_chunks([CommandOutputChunk::Stderr(Bytes::from_static(b"err"))]);
+
+        assert_eq!(ToolOutput::from(output), ToolOutput::Text("err".into()));
+    }
+}
+
 impl std::fmt::Display for ToolOutput {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
--- a/swiftide-core/src/chat_completion/tools.rs
+++ b/swiftide-core/src/chat_completion/tools.rs
@@ -113,32 +113,6 @@ impl From<CommandOutput> for ToolOutput {
         ToolOutput::Text(output.into_string_lossy())
     }
 }
-
-#[cfg(test)]
-mod command_output_tests {
-    use super::{CommandOutput, ToolOutput};
-    use crate::CommandOutputChunk;
-    use bytes::Bytes;
-
-    #[test]
-    fn keeps_stdout_and_stderr_in_tool_output() {
-        let output = CommandOutput::from_chunks([
-            CommandOutputChunk::Stdout(Bytes::from_static(b"out")),
-            CommandOutputChunk::Stderr(Bytes::from_static(b"err")),
-        ]);
-
-        assert_eq!(ToolOutput::from(output), ToolOutput::Text("outerr".into()));
-    }
-
-    #[test]
-    fn keeps_stderr_only_tool_output() {
-        let output =
-            CommandOutput::from_chunks([CommandOutputChunk::Stderr(Bytes::from_static(b"err"))]);
-
-        assert_eq!(ToolOutput::from(output), ToolOutput::Text("err".into()));
-    }
-}
-
 impl std::fmt::Display for ToolOutput {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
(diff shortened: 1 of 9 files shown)


swiftide-core/src/chat_completion/tools.rs
```diff
@@ -110,7 +110,7 @@ impl<S: AsRef<str>> From<S> for ToolOutput {
 
 impl From<CommandOutput> for ToolOutput {
     fn from(output: CommandOutput) -> Self {
-        ToolOutput::Text(output.into_string_lossy())
+        ToolOutput::Text(output.into_stdout_string_lossy())
```
timonv: The default should be into_string_lossy still

swiftide-core/src/command_output.rs
```diff
@@ -51,48 +64,70 @@ impl<T: Into<Vec<u8>>> From<T> for CommandOutput {
 
 impl std::fmt::Display for CommandOutput {
     fn fmt(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        formatter.write_str(&self.to_string_lossy())
+        if self.stdout.is_empty() {
+            formatter.write_str(&self.stderr_to_string_lossy())
+        } else {
+            formatter.write_str(&self.stdout_to_string_lossy())
+        }
```
timonv: This should just do self.to_string_lossy() like what it was

swiftide-core/src/command_output.rs
```diff
@@ -1,49 +1,237 @@
...
+        let mut bytes = Vec::with_capacity(
+            self.selected_chunks(kind)
+                .map(CommandOutputChunk::as_bytes)
+                .map(<[u8]>::len)
+                .sum(),
+        );
+        for chunk in self.selected_chunks(kind) {
+            bytes.extend_from_slice(chunk.as_bytes());
+        }
+        Cow::Owned(bytes)
```
timonv: I wonder if this function can just return:

fn selected_bytes(&self, kind: Option<OutputKind>) -> Bytes
self.selected_chunks(kind)
                .map(CommandOutputChunk::as_bytes)
                .collect()
}

Otherwise it's essentially always owned, reallocating every command output.

I think in general we should remove the Cow (rest as well) and aim for it to be borrowed bytes by default. Otherwise it becomes hard to say what is owned and what is borrowed

swiftide-core/src/command_output.rs
```diff
@@ -1,49 +1,237 @@
...
+        }
+        bytes
     }
 }
 
-impl<T: Into<Vec<u8>>> From<T> for CommandOutput {
+/// A borrowed view of one command output pipe.
+///
+/// Obtain a view with [`CommandOutput::stdout`] or [`CommandOutput::stderr`]. It references the
+/// ordered chunk list and does not create a second stdout or stderr buffer.
```
timonv: This is not true, most cases it will be Cow::Owned

swiftide-core/src/command_output.rs
timonv: Fixed. Display now delegates to to_string_lossy().

swiftide-core/src/command_output.rs
timonv: Fixed by removing selected_bytes, CommandOutputView, OutputKind, and the public Cow return values. stdout() and stderr() now iterate borrowed byte slices directly. Combined text conversion returns an owned String.

swiftide-core/src/command_output.rs
timonv: Fixed. The Cow-based view and its ownership claims are removed; the stream methods now expose borrowed byte slices directly.

swiftide-core/src/command_output.rs
```diff
@@ -1,49 +1,142 @@
...
+pub enum CommandOutputChunk {
+    /// Bytes read from stdout.
+    Stdout(Bytes),
+    /// Bytes read from stderr.
+    Stderr(Bytes),
+}
+
+impl CommandOutputChunk {
+    /// Returns the bytes in this chunk.
+    pub fn as_bytes(&self) -> &[u8] {
```
timonv: Should it return &Bytes?

swiftide-core/src/command_output.rs
```diff
@@ -1,49 +1,142 @@
...
+    /// Iterates over borrowed stdout chunks in observed order.
+    pub fn stdout(&self) -> impl Iterator<Item = &[u8]> {
+        self.chunks.iter().filter_map(|chunk| match chunk {
+            CommandOutputChunk::Stdout(bytes) => Some(bytes.as_ref()),
+            CommandOutputChunk::Stderr(_) => None,
+        })
+    }
+
+    /// Iterates over borrowed stderr chunks in observed order.
+    pub fn stderr(&self) -> impl Iterator<Item = &[u8]> {
```
timonv: Shouldn't this and stdout iterate on &Bytes?

swiftide-core/src/command_output.rs
```diff
@@ -1,49 +1,142 @@
...
+
+    /// Converts all output to owned text in observed chunk order.
+    pub fn to_string_lossy(&self) -> String {
+        bytes_into_string_lossy(self.bytes())
     }
 
-    pub fn to_string_lossy(&self) -> Cow<'_, str> {
-        String::from_utf8_lossy(&self.bytes)
+    /// Returns all output bytes in observed chunk order.
+    pub fn into_bytes(self) -> Vec<u8> {
```
timonv, line 102: Shouldn't this be Vec<Bytes>

Any why can't this method be just: `self.chunks.into_iter().collect()` ?

swiftide-core/src/command_output.rs
```diff
@@ -55,44 +148,103 @@ impl std::fmt::Display for CommandOutput {
     }
 }
 
+fn bytes_into_string_lossy(bytes: Vec<u8>) -> String {
```
timonv, line 143: This should return a Cow<_,str> so that the latter `into_owned` is not needed

swiftide-core/src/command_output.rs
timonv: Done. `CommandOutputChunk::as_bytes` now returns `&Bytes`.

swiftide-core/src/command_output.rs
timonv: Done. `stdout()` and `stderr()` now iterate over `&Bytes`.

swiftide-core/src/command_output.rs
timonv: Kept this as one combined `Vec<u8>` to preserve the existing consuming API. It reserves once and copies each chunk once. Callers that want the existing chunks without joining can use `chunks()` or consume the chunk list through the type itself; changing this method to `Vec<Bytes>` would change its meaning and every combined-output caller.

swiftide-core/src/command_output.rs
timonv: Done. `to_string_lossy()` now returns `Cow<'_, str>` and borrows when the combined bytes are already borrowed.

PR #1170, opened 2026-09-18
refactor(core): fold buffered output reporting into the sink trait

timonv:
## What changed

Follow-up to #1162, after review of the streaming code for further reductions.

- `report_buffered_output` is gone as a public free function. It is now a defaulted method on `CommandOutputSink`, so callers reach it from the sink they already hold.
- The unit sink impl (`()`) is gone. The one caller passes an empty closure instead.
- `LocalExecutor::exec_cmd_streaming` returns early for shell commands and wraps the buffered file arms once, so the "shell streams, files report afterwards" shape is visible.
- `exec_cmd` no longer opens its own tracing span on top of the one in `exec_cmd_streaming`. The remaining span records the command.
- The sink trait documents that `on_chunk` runs inside the read loop and must not block.

## Why

One public item instead of two for the same concept, and less code overall for the same behaviour.

## Checks

- Clippy and rustfmt clean.
- All tests in swiftide-core and swiftide-agents pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

- **New Features**
  - Command execution now supports streaming output as it becomes available.
  - Output is delivered in order for successful commands, failures, and timed-out processes.
  - Buffered command results and startup errors are also reported through the output stream.
  - Existing command execution behavior remains available.


timonv:
Superseded by #1171, which carries the same commit rebased onto master.

